PR #6, opened 2019-03-25
Add email sending address setting

joehoyle:
Also provided this with a default so atleast we'll be able to send email when this is not configured

### DIFF
--- a/docs/email-delivery.md
+++ b/docs/email-delivery.md
@@ -1,0 +1,43 @@
+# Email Delivery
+
+Platform Cloud can handle vast amounts of email sending, with very good deliver-ability. Email sending is automatically integrated via a custom `wp_mail` override, therefore developers and plugins don't need to make any extra provisions for sending email on Cloud. This integration is controlled via `aws-ses`.
+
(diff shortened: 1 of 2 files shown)


docs/email-delivery.md
```diff
@@ -0,0 +1,43 @@
...
+}
+```
+
+For advanced configuration of the sending email address for outbound mail, use the `wp_mail_from` filter to change the value at runtime:
+
+```php
+add_filter( 'wp_mail_from', function ( $email ) {
+	return 'no-reply@mydomainname.com';
+} );
+
```
roborourke: Bonus newline

docs/email-delivery.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+To configure the "from name" of outbound emails by using the `wp_mail_from_name` filter:
+
+```php
+add_filter( 'wp_mail_from_name', function ( $name ) {
+	return 'Joe Bloggs';
+} );
+
```
roborourke: Bonus newline

PR #7, opened 2019-04-02
Turn module on for all environments

rmccue:
This changes the configuration to be more granular, and only turns off the non-production parts of the configuration. This ensures things like Redis work locally.

### DIFF
--- a/load.php
+++ b/load.php
@@ -11,7 +11,7 @@ use HM\Platform\XRay;
 require_once __DIR__ . '/inc/namespace.php';
 
 add_action( 'hm-platform.modules.init', function () {
-	$is_cloud = in_array( get_environment_architecture(), [ 'ec2', 'ecs' ] );
+	$is_cloud = in_array( get_environment_architecture(), [ 'ec2', 'ecs' ], true );
 	$default_settings = [
 		'enabled'            => true,
 		'cavalcade'          => true,
--- a/load.php
+++ b/load.php
@@ -11,16 +11,17 @@ use HM\Platform\XRay;
 require_once __DIR__ . '/inc/namespace.php';
 
 add_action( 'hm-platform.modules.init', function () {
+	$is_cloud = in_array( get_environment_architecture(), [ 'ec2', 'ecs' ] );
 	$default_settings = [
-		'enabled'            => in_array( get_environment_architecture(), [ 'ec2', 'ecs', 'local-server' ] ),
+		'enabled'            => true,
 		'cavalcade'          => true,
 		's3-uploads'         => true,
-		'aws-ses-wp-mail'    => true,
-		'batcache'           => true,
+		'aws-ses-wp-mail'    => $is_cloud,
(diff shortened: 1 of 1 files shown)


joehoyle:
Do we want redis locally? I thought Chassis used memcached for example. it seems these things are enviornment specific. Cloud, local-server and local-server-chassis are "environment modules" so I think it might make more sense to keep them mutually exclusive

rmccue:
Chassis can use whatever you want, doesn't include anything out of the box. The setup script includes Redis though.

The problem is that we're going to end up with a lot of duplication if we separate fully by environment. Right now, local-server is covered by Cloud, so may as well add Chassis to that list.

load.php
```diff
@@ -11,16 +11,17 @@
...
 		's3-uploads'         => true,
-		'aws-ses-wp-mail'    => true,
-		'batcache'           => true,
+		'aws-ses-wp-mail'    => $is_cloud,
+		'batcache'           => $is_cloud,
 		'redis'              => true,
 		'ludicrousdb'        => true,
 		'healthcheck'        => true,
-		'xray'               => true,
+		'xray'               => $is_cloud,
```
joehoyle: We use / support this in local-server, so might need a way to modify this one

load.php
```diff
@@ -11,16 +11,17 @@
...
+	$is_cloud = in_array( get_environment_architecture(), [ 'ec2', 'ecs' ], true );
 	$default_settings = [
-		'enabled'            => in_array( get_environment_architecture(), [ 'ec2', 'ecs', 'local-server' ] ),
+		'enabled'            => true,
 		'cavalcade'          => true,
 		's3-uploads'         => true,
-		'aws-ses-wp-mail'    => true,
-		'batcache'           => true,
+		'aws-ses-wp-mail'    => $is_cloud,
+		'batcache'           => $is_cloud,
```
joehoyle: I'm wondering if it would be better to put this on anyway? Having page caching locally is annoying when you don't want it, but we probably do want developers to be maybe be getting it by default?

load.php
rmccue: Interesting; using a local XRay, or pushing to live? We should try and support in Chassis if it's not too hard.

load.php
rmccue: I'd say no, you can always config it on if needed.

load.php
joehoyle: The xray daemon runs locally but then pushed to aws. You need aws creds currently to do that. It's mostly used for developing the xray plugin its self.

load.php
roborourke: I'm generally ok with it being on as you rarely generate enough traffic to trigger batcache locally unless it's intentional. I prefer it on personally as it can show potential issues with batcache locally however I'm fine if it's something we just need to turn on manually as it means you've done so mindfully and intentionally.

rmccue: I'd say better to be explicit about turning it on, as debugging it otherwise can be a total pain

PR #9, opened 2019-04-03
Define TACHYON_URL in cloud module

joehoyle:
As per the code comment, we don't define this in the infra in ECS or EC2

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -183,6 +183,14 @@ function load_plugins() {
 		define( 'DISABLE_WP_CRON', true );
 	}
 
+	// Define TACHYON_URL, as in the Cloud environment is "always on"
+	// but the constant is not defined at the infra. level as we want
+	// it to be the network primary domain which isn't available
+	// at the infra level current.
+	if ( ! defined( 'TACHYON_URL' ) ) {
+		define( 'TACHYON_URL', get_site_url( get_main_site_id( get_main_network_id() ), '/tachyon' ) );
+	}
+
 	foreach ( get_available_plugins() as $plugin => $file ) {
 		if ( ! $config[ $plugin ] ) {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -183,6 +183,14 @@ function load_plugins() {
 		define( 'DISABLE_WP_CRON', true );
 	}
 
+	// Define TACHYON_URL, as in the Cloud environment is "always on"
+	// but the constant is not defined at the infra. level as we want
+	// it to be the network primary domain which isn't available
+	// at the infra level current.
+	if ( ! defined( 'TACHYON_URL' ) ) {
+		define( 'TACHYON_URL', get_site_url( get_main_site_id( get_main_network_id() ), '/tachyon' ) );
```
roborourke: This way we might still have images being served from a different domain than the current one. This makes sense from a caching point of view and perhaps not a problem but depending on the CORS headers we're providing could cause issues with loading in images via JS eg. in the image cropping UI for smart media.

Will any domain used by subsites have the `/tachyon` path or only the primary one?

joehoyle: Hmm, how are they loaded in JS? I thought that would still work. Any domain will have the tachyon path available yeah

roborourke: I definitely ran into the issue building smart media, image tags are added and removed via backbone templates. Will probably be quickest to confirm with chassis.

joehoyle: I think either way, we are not doing anything worse than what we have now here. TO my knowledge we don't define TACHYON_URL per-domain on any other sites. If we _do_ need to do that, I think it can be a follow-up as I don't think we've solved this problem before this patch anyway?

roborourke: Yep, agreed.

PR #11, opened 2019-04-08
Allow overriding Accept with `_accept=json`

joehoyle:
Currently we don't pass the `Accept` header to the origin via the CDN, for cache performance reasons. This is unlikely to change soon, so I added support for specifying it via a query param.

### DIFF
--- a/plugins/healthcheck/inc/namespace.php
+++ b/plugins/healthcheck/inc/namespace.php
@@ -31,7 +31,15 @@ function output_page( array $checks ) {
 	}
 	nocache_headers();
 
+	$format = 'html';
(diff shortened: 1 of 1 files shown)


nathanielks:
Wouldn't passing `Accept` have a negligible difference, and wouldn't passing a query param have the same effect? Or is there a significant diversity in what browsers submit as an `Accept` header? I don't like the idea of a query param and would strongly prefer the Accepts header.

joehoyle:
@nathanielks yes, that's the problem - there's a high degree of variance in the Accept header from browser to browser. It's why we don't use it in the cache key now, and also why we don't use it in Tachyon, (instead using lambda@edge to determine webp support based off the accept header)

nathanielks:
Understood. And the reason it won't make a significant difference via a query parameter is because we're manually specifying it, so it's not going to make a significant difference 👍

PR #20, opened 2019-04-19
Remove install plugins / themes in Cloud environments

joehoyle:
I think for subtle reasons, and practically, it's better not to have these available.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -60,6 +60,10 @@ function bootstrap( $wp_debug_enabled ) {
 	add_filter( 'enable_loading_advanced_cache_dropin', __NAMESPACE__ . '\\load_advanced_cache', 10, 1 );
 	add_action( 'muplugins_loaded', __NAMESPACE__ . '\\load_plugins' );
 
+	if ( in_array( get_environment_architecture(), [ 'ec2', 'ecs' ], true ) ) {
+		add_filter( 'map_meta_cap', __NAMESPACE__ . '\\disable_install_capability', 10, 2 );
+	}
+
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -216,3 +220,20 @@ function load_plugins() {
 		require dirname( __DIR__ ) . '/plugins/' . $file;
 	}
 }
+
+/**
+ * Disable the install plugins / themes capability in Cloud, as it's not possible to do in that context.
+ *
+ * @param array $caps
+ * @param string $cap
```
rmccue: As @roborourke mentioned, these should be documented.

inc/namespace.php
```diff
@@ -216,3 +220,20 @@ function load_plugins() {
 		require dirname( __DIR__ ) . '/plugins/' . $file;
 	}
 }
+
+/**
+ * Disable the install plugins / themes capability in Cloud, as it's not possible to do in that context.
```
rmccue: This is too long, split to short description + long desc.

roborourke:
```suggestion
 * Disable the install plugins / themes capability.
 * 
 * In the cloud context it's not possible to do this so this hides 
 * non functional UI.
```

inc/namespace.php
```diff
@@ -216,3 +220,20 @@ function load_plugins() {
...
+ *
+ * @param array $caps
```
roborourke:
```suggestion
 * @param array $caps The required capabilities list.
```

inc/namespace.php
roborourke:
```suggestion
 * @param string $cap The current capability being checked.
```

PR #33, opened 2019-04-30
Add docs on HTTPS

joehoyle:


joehoyle:
Yeah that will change with the rename so I think it's fine

PR #36, opened 2019-05-10
Cache admin-ajax requests for logged out users

### LINKED ISSUE #34: Forward port batcache admin-ajax
Added to hm-platform in https://github.com/humanmade/hm-platform/pull/161

joehoyle:
I thought this was already the case, turns out it is not. `admin-ajax.php` is often used as a quick HTTP API to power the front-end of the site for features like pagination, and we want to be caching all those requests like any other.

Fixes #34.


joehoyle:
> Also I'm assuming that adding the action after return $should_load is intentional?

Hmm yeah technically I wouldn't need to be, but then I wouldn't put it in this function at all.

PR #45, opened 2019-05-24
Fix coding standards

roborourke:


inc/cavalcade_runner_to_cloudwatch/namespace.php
```diff
@@ -1,19 +1,23 @@
...
 use ReflectionClass;
-use function Altis\get_aws_sdk;
 
-Runner::instance()->hooks->register( 'Runner.run_job.started', __NAMESPACE__ . '\\on_job_started' );
-Runner::instance()->hooks->register( 'Runner.check_workers.job_failed', __NAMESPACE__ . '\\on_job_failed' );
-Runner::instance()->hooks->register( 'Runner.check_workers.job_completed', __NAMESPACE__ . '\\on_job_completed' );
+/**
+ * Register Cavalcade Hooks.
+ */
+function bootstrap() {
```
rmccue: This will still run in the Cavalcade Runner context right?

inc/namespace.php
```diff
@@ -3,14 +3,50 @@
...
+		&& ( ! defined( 'WP_CLI' ) || ! WP_CLI )
+		&& ! class_exists( 'HM\\Cavalcade\\Runner\\Runner' )
+	) {
+		require_once ROOT_DIR . '/vendor/humanmade/aws-xray/inc/namespace.php';
+		require_once ROOT_DIR . '/vendor/humanmade/aws-xray/plugin.php';
+		XRay\bootstrap();
+	}
+
+	if ( $config['batcache'] && ! defined( 'WP_CACHE' ) ) {
+		// @codingStandardsIgnoreLine
```
rmccue: Does this have to be ignored now that it's in namespace.php?

inc/cavalcade_runner_to_cloudwatch/namespace.php
roborourke: Yep, I added the call to it after the file is included here https://github.com/humanmade/altis-cloud/pull/45/files#diff-2d9bc91244a2cb57445498ad1c971f7dR48

PR #46, opened 2019-05-29
Add docs on static file caching

joehoyle:


docs/static-file-caching.md
```diff
@@ -0,0 +1,13 @@
...
+
+Because query strings are part of the cache key when the file is cached on the CDN, in many cases it's simplest to reference the file with a changed `?version` query parameter. This behavior is automatic when providing the `$version` parameter to the `wp_enqueue_script` and `wp_enqueue_style` functions.
+
+```php
+wp_enqueue_script( 'my-theme', get_stylesheet_uri(), [], '2019-05-4-1' );
+```
+
+This will enqueue and output a `<script>` tag with a URL appended with `?ver=2019-05-4-1`. Every time the `style.css` is changed, the `$version` parameter needs to be updated.
+
+The `$version` parameter can be defined dynamically, though should be optimized to only change when the file has changed. To reduce disk IO, you should not read the file at runtime to generate a version, such as `filemtime` or `md5_file`. Using Webpack or other build-time workflows to build assets on deploy will provide unique filenames that can provide hard coded values.
```
roborourke: Maybe break this bit up by adding a heading like "Performance considerations" to help it stand out a bit.

docs/static-file-caching.md
```diff
@@ -0,0 +1,13 @@
...
+
+All static files on Altis Cloud are cached on the CDN for a long period of time. This ensures the best performance when downloading static assets such as images, JavaScript and CSS files. Due to the aggressive caching times, when files are changed it is necessary reference the files by a new URL that will cause a different version to be saved on the CDN.
+
+Because query strings are part of the cache key when the file is cached on the CDN, in many cases it's simplest to reference the file with a changed `?version` query parameter. This behavior is automatic when providing the `$version` parameter to the `wp_enqueue_script` and `wp_enqueue_style` functions.
+
+```php
+wp_enqueue_script( 'my-theme', get_stylesheet_uri(), [], '2019-05-4-1' );
+```
+
+This will enqueue and output a `<script>` tag with a URL appended with `?ver=2019-05-4-1`. Every time the `style.css` is changed, the `$version` parameter needs to be updated.
```
roborourke: Could mention the deployment revision constant too? Currently `HM_DEPLOYMENT_REVISION`.

docs/static-file-caching.md
```diff
@@ -0,0 +1,13 @@
+# Static File Caching
+
+All static files on Altis Cloud are cached on the CDN for a long period of time. This ensures the best performance when downloading static assets such as images, JavaScript and CSS files. Due to the aggressive caching times, when files are changed it is necessary reference the files by a new URL that will cause a different version to be saved on the CDN.
```
roborourke: Can/should we give the precise times here?

docs/static-file-caching.md
joehoyle: Hmm I think it's better as is, it's not so much a consideration rather disallowed.

docs/static-file-caching.md
joehoyle: Yeah I was going to, but then realized we don't have a better altis name for it. I think we should have another helper function ala `Altis\get_environment_architecture()` rather than the constant. Created https://github.com/humanmade/altis-core/issues/34 to track

docs/static-file-caching.md
joehoyle: I think it's ok to just be long-term. If we want to change the specific time that gives us more freedom. From a n=dev perspective the important thing is "it's a long time"

PR #49, opened 2019-06-02
Ensure Cloud configuration is respected.

peterwilsoncc:
Adds checks to ensure the following configuration options are respected:

* `ludicrousdb`
* Either `redis` or `memcached` are enabled before loading object caching
* Logic check of `email-from-address` to ensure it's set and not falsey

Additionally I am unsure whether 

* `bootstrap()` should return early if `! $config['enabled']`
* `SES_To_CloudWatch\bootstrap();` and `CloudWatch_Logs\bootstrap();` should be conditional on `audit-log-to-cloudwatch || php-errors-to-cloudwatch`.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -137,6 +137,34 @@ function get_config() {
 }
 
 /**
+ * Load the object cache.
+ *
+ * Check the object caching configuration and load either memcached
+ * or redis as appropriate.
+ *
+ * @deprecated 1.1.0 Object caching setup moved to dedicated functions.
+ */
+function load_object_cache() {
+	if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+		trigger_error(
+			sprintf(
+				__( '%1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.' ),
+				__FUNCTION__,
+				'1.1.0',
+				'load_object_cache_*()'
+			)
+		);
+	}
+	$config = get_config();
+
+	if ( $config['memcached'] ) {
+		load_object_cache_memcached();
+	} elseif ( $config['redis'] ) {
+		load_object_cache_redis();
+	}
+}
+
+/**
  * Load the Memcached Object Cache dropin.
  */
 function load_object_cache_memcached() {
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -70,8 +70,15 @@ function load_platform( $wp_debug_enabled ) {
 		$_SERVER['HTTP_USER_AGENT'] = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.86 Safari/537.36';
 	}
 
-	load_object_cache();
-	load_db();
+	if ( $config['memcached'] ) {
+		load_object_cache_memcached();
+	} elseif ( $config['redis'] ) {
+		load_object_cache_redis();
+	}
+
+	if ( $config['ludicrousdb'] ) {
+		load_db();
+	}
 
 	global $wp_version;
 	if ( version_compare( '4.6', $wp_version, '>' ) ) {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -29,7 +29,7 @@ function bootstrap() {
 	}
 
 	add_filter( 'wp_mail_from', function ( string $email ) use ( $config ) : string {
-		return $config['email-from-address'];
+		return ! empty( $config['email-from-address'] ) ? $config['email-from-address'] : $email;
```
roborourke: We could probably validate this too with a filter function like:
```suggestion
		return filter_var(
			$config['email-from-address'],
			FILTER_VALIDATE_EMAIL,
			FILTER_NULL_ON_FAILURE
		) ?? $email;
```

inc/namespace.php
```diff
@@ -133,8 +133,11 @@ function get_config() {
  * Load the Object Cache dropin.
  */
 function load_object_cache() {
-	wp_using_ext_object_cache( true );
 	$config = get_config();
+	if ( ! $config['memcached'] && ! $config['redis'] ) {
```
roborourke: I think I would move these checks to where the functions are called in `load_platform()`. Looking at further now I'd probably also split the redis and memcached loading functions so that they just have one task each and don't need to check the config. That would make the functions behave the same way whenever they're called which will simplify any future unit tests.

inc/namespace.php
```diff
@@ -187,6 +190,10 @@ function disable_no_cache_headers_on_admin_ajax_nopriv() {
  * Load the db dropin.
  */
 function load_db() {
+	$config = get_config();
```
roborourke: So as per the above comment, I think this should go in the `load_platform()` function.

inc/namespace.php
peterwilsoncc: I've done this in e8046bc2df8b733c678d63abdeb16c93bb2a7afa.

To keep with back-compat and LTS policies, I've added `load_object_cache()` back in to trigger a notice in 30804921d875fd7ef5624a732fa51d532d6552d3.

inc/namespace.php
peterwilsoncc: Also done in e8046bc2df8b733c678d63abdeb16c93bb2a7afa

inc/namespace.php
```diff
@@ -130,26 +141,59 @@ function get_config() {
...
+ * or redis as appropriate.
+ *
+ * @deprecated 1.1.0 Object caching setup moved to dedicated functions.
  */
 function load_object_cache() {
-	wp_using_ext_object_cache( true );
+	if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+		trigger_error(
+			sprintf(
+				__( '%1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.' ),
```
peterwilsoncc: This is a core translation string so I've not put the `altis` domain in.

I am not sure whether the i18n are available at this point so it may need to be removed.

roborourke: @peterwilsoncc good call checking on that - the translation functions aren't available at that time, it's fine for this to be an untranslated string as it's not user facing.

roborourke:
```suggestion
				'%1$s is deprecated since version %2$s! Use %3$s instead.',
```

inc/namespace.php
```diff
@@ -130,26 +141,59 @@ function get_config() {
...
+				__FUNCTION__,
+				'1.1.0',
```
roborourke: I want to double check this but pretty sure we won't be doing any 1.x.x releases that aren't RC versions. Because it's a bug fix and backwards compatible we'll do this as a patch release.
```suggestion
				'1.0.1',
```

inc/namespace.php
```diff
@@ -130,26 +141,59 @@ function get_config() {
 }
 
 /**
- * Load the Object Cache dropin.
+ * Load the object cache.
+ *
+ * Check the object caching configuration and load either memcached
+ * or redis as appropriate.
+ *
+ * @deprecated 1.1.0 Object caching setup moved to dedicated functions.
```
roborourke:
```suggestion
 * @deprecated 1.0.1 Object caching setup moved to dedicated functions.
```

PR #52, opened 2019-06-17
Document how to configure nginx to block a file

kadamwhite:
This could be reorganized to have more document hierarchy later on, but this example came up in Slack and having it spelled out explicitly may help future Altis developers. Snippet kindly provided by @nathanielks

### DIFF
--- a/docs/nginx-configuration.md
+++ b/docs/nginx-configuration.md
@@ -16,10 +16,10 @@ if ( $host = "example.com" ) {
 }
 ```
 
-Alternatively, if you have a local file in your repository you wish to explicitly block from being publicly accessed you may manually configure a 404 for a specific resource or filename pattern:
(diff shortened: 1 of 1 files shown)


joehoyle:
Wondering if this will block the root file by that name, or all files in any dir of that name? Or -- which of those do you intend?

nathanielks:
This should block all files with that name. Blocking the root would need a forward slash at the beginning. I gave that example, so I was assuming it'd be good to block all occurances!

joehoyle:
Ok, I think we might be able to make the example description clearer about that, currently it says "if you have a local file in your repository you wish to explicitly block"

kadamwhite:
@joehoyle Revised the prose & comment. Look OK?

joehoyle:
Great stuff, thanks @kadamwhite

PR #54, opened 2019-07-02
Fix typo in page cache cookie example and change HM reference to Altis support

### LINKED ISSUE #53: Docs: typo and reference to HM
On this page there's a typo in the page cache cookie example and also a reference to requesting support for specific cookies from Human Made which should be updated to Altis support.

https://www.altis-dxp.com/resources/docs/cloud/page-caching/

joeleenk:
• Fix typo
• Switch HM reference to Altis support

Closes #53

### DIFF
--- a/docs/page-caching.md
+++ b/docs/page-caching.md
@@ -33,20 +33,20 @@ Any response that can be added to the page cache should not include references t
 
 ## Cache Rule Customizations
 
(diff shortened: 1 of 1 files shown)


joehoyle:
@joeleenk just a reminder that we use the PR titles in the changelog, so it’s best to title the pr accordingly. In this case: I’d not I close the issue number, and be a bit more specific in the change

joeleenk:
@joehoyle Thanks for the info! I updated the title of the PR - is that more what you're looking for?

PR #59, opened 2019-08-02
Fix branding

glueckpress:
This PR eliminates all appearances of the string `HM Cloud` (case-sensitive) in favour of `Altis`.

Didn’t touch any code; docs and 1 comment in wp-config.php only.


glueckpress:
Lol, sorry about the funny branch name, somebody was clearly over-achieving there. 😬

PR #61, opened 2019-08-20
Update plugin to adhere to humanmade/coding-standards

nathanielks:
Before I get to work on fixing the negative maxage bug we've encountered, I wanted to update the repo to be compliant with hm-linter. This updates the plugin to adhere to our coding standards, except in instances where it was simpler to ignore to maintain back-compat: Namespaces and changing the name of the batcache class.

### DIFF
--- a/dropins/batcache/advanced-cache.php
+++ b/dropins/batcache/advanced-cache.php
@@ -1,389 +1,30 @@
 <?php
-if ( is_readable( dirname( __FILE__ ) . '/batcache-stats.php' ) )
-	require_once dirname( __FILE__ ) . '/batcache-stats.php';
 
-if ( !function_exists( 'batcache_stats' ) ) {
-	function batcache_stats( $name, $value, $num = 1, $today = FALSE, $hour = FALSE ) { }
+if ( is_readable( dirname( __FILE__ ) . '/batcache-stats.php' ) ) {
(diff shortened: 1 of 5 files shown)


nathanielks:
Ah, I now see we include batcache directly instead of using our fork of the repo. I also see we ignore this whole directory from linting. I can still see value in adding these updates as my editor shows linting errors in my sidebar. It also makes it easier to fix formatting errors because I can run `phpcbf` on the whole file without worrying about breaking something.

joehoyle:
I think instead of this we should move these to composer per #5, then (if we want) file these changes upstream. We’ll prob Amy need to add our forks to composer too to switch these to composer deps

nathanielks:
@joehoyle 	got it 👍 	For clarification, can you clarify what you mean by "We’ll [probably] need to add our forks to composer too to switch these to composer deps?"  I'm not 100% sure how forks play into this. My understanding of what needs to happen is we need to take the packages currently in `plugins/` and `dropins/` and move them into their own repository (if it doesn't exist), add composer support, then publish on Packagist.

joehoyle:
Yup exactly, though we already have repos for these: https://github.com/humanmade/batcache

nathanielks:
Got it, thanks! I'll work on doing that 👍

PR #65, opened 2019-09-12
Set the DB `dbname` etc properties

joehoyle:
LudicrousDB doesn't provide the `dbname`, `dbuser` etc instance vars on the `global $wpdb` object, which is needed by WP CLI (and some others) to detect what the DB name is. This is used when getting a list of tables for example, which breaks with LudicrousDB. Also, search-replace commands also break unless passed the `--all-tables` flag.

LudicrousDB doesn't set these almost for a good reason: each host you add to LudicrousDB _could_ use a different database name, so there is no "single" DB name, hence no setting of an instance variable for `dbname`. Instead, I decided to go for a better default: if a database host is configured that has `write` properties, then use that one as the host (etc) references in the instance variables.

### DIFF
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -22,7 +22,6 @@ class DB extends LudicrousDB {
 	 */
 	public $time_spent = 0;
 
-
 	function query( $query ) {
 		$start = microtime( true );
 		$result = parent::query( $query );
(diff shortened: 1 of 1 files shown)


rmccue:
I'd be up for us backporting.

PR #66, opened 2019-09-13
Add some pre-configuration for batcache

roborourke:
This is a pre-cursor to using the `CloudFront-Viewer-Country` in response headers for analytics. I added some good defaults in too.

- CloudFront header reflection
- Default batcache ignored query params (configurable)
- Configurable batcache unique keys based on headers
- Configurable batcache unique keys based on cookies

Not sure if this is the right approach exactly, especially with regards to forwarding the headers back.

Things still to add:

- ~[ ] helper functions based on CF headers eg. `is_country()`, `is_mobile()`~
- [x] Updated docs


inc/namespace.php
```diff
@@ -212,6 +212,10 @@ function load_advanced_cache( $should_load ) {
 	}
 	add_action( 'admin_init', __NAMESPACE__ . '\\disable_no_cache_headers_on_admin_ajax_nopriv' );
 
+	// Pre-configure batcache.
+	require __DIR__ . '/page_cache/namespace.php';
+	Page_Cache\bootstrap();
```
joehoyle: I think we should probably move the batcache loading into this part too?

inc/page_cache/namespace.php
```diff
@@ -0,0 +1,60 @@
...
+	global $batcache;
+	$batcache = $batcache ?? [];
+
+	$config = get_config();
+	$unique_headers = $config['cache']['unique-headers'] ?? [
+		'CloudFront-Is-Desktop-Viewer',
+		'CloudFront-Is-Mobile-Viewer',
+		'CloudFront-Is-SmartTV-Viewer',
+		'CloudFront-Is-Tablet-Viewer',
+		'CloudFront-Viewer-Country',
```
joehoyle: I think by default we don't want to make country (or maybe any?) of these part of the `unique` key? For example, I think you wanted this for analytics, whereby we wanted to echo back the `CloudFront-Viewer-Country` _pre_ batcache, i.e. not have a page cache per country.

inc/page_cache/namespace.php
```diff
@@ -0,0 +1,60 @@
...
+		'utm_content',
+		'fbclid',
+		'_ga',
+	];
+
+	$batcache['ignored_query_string_params'] = $ignored_query_string_params;
+}
+
+function handle_cloudfront_headers() {
+	global $batcache;
```
joehoyle: I think rather than having these functions be changing the global, I'd have them maybe be `get_batcache_config_for_cloudfront_headers` type thing, then have  `bootstrap` or similar handle merging those configs into the batcache global. Otherwise, we have a bunch of nasty side effects functions in this NS.

inc/page_cache/namespace.php
```diff
@@ -0,0 +1,60 @@
...
+	// Forward CloudFront cache headers.
+	handle_cloudfront_headers();
+}
+
+function ignore_query_params() {
+	global $batcache;
+	$batcache = $batcache ?? [];
+
+	$config = get_config();
+	$ignored_query_string_params = $config['cache']['ignored-query-string-params'] ?? [
```
joehoyle: Could this be part of the defaults in the module?

roborourke: Yeah, wasn't sure of the overall approach yet so just put them there for testing.

inc/page_cache/namespace.php
roborourke: Right I see, was thinking that was tied to the page cache too for some reason but that's just at the CloudFront level. So those headers can be reflected back regardless if they're set - the config can be used to opt them into the `unique` key.

This will require an infrastructure change too so I'll raise an issue for that before we go any further here.

inc/page_cache/namespace.php
roborourke: Yep that makes sense! That way we can add a guard to check if it's appropriate to run in a single location if it's called again for any reason.

roborourke:
@joehoyle made some updates, think this a little bit more in line with what you were saying. Will make sure that it doesn't clobber any existing array values too.

roborourke:
@joehoyle / @rmccue any chance of another look over this? I feel like I'm missing something

load.php
```diff
@@ -15,19 +15,31 @@
...
+		'batcache' => $is_cloud,
+		'redis' => true,
+		'memcached' => get_environment_architecture() === 'ec2',
+		'ludicrousdb' => true,
+		'healthcheck' => true,
+		'xray' => $is_cloud,
+		'email-from-address' => 'no-reply@humanmade.com',
+		'audit-log-to-cloudwatch' => $is_cloud,
 		'php-errors-to-cloudwatch' => $is_cloud,
+		'cache' => [
```
joehoyle: I think this should be called `page-cache` maybe?

inc/namespace.php
```diff
@@ -322,3 +331,28 @@ function disable_install_capability( array $caps, string $cap ) : array {
...
+ * This function reflects those headers back in responses for
+ * client side usage.
+ */
+function reflect_cloudfront_headers() {
+	$headers = [
+		'CloudFront-Is-Desktop-Viewer',
+		'CloudFront-Is-Mobile-Viewer',
+		'CloudFront-Is-SmartTV-Viewer',
+		'CloudFront-Is-Tablet-Viewer',
+		'CloudFront-Viewer-Country',
```
joehoyle: Apart from the Viewer Country, I'm wondering if we should add these. I don't think we'd ever configure cloudfront to segment caches based off `CloudFront-Is-Tablet-Viewer`, as we'd expect that detection to be done on the client. Are there are of these apart from country that can not be done on the client-side?

load.php
roborourke: Yeah makes sense to be more specific 👍

inc/namespace.php
roborourke: So this only reflects the headers if they're set. I'm not suggesting we enable any of the others besides `CloudFront-Viewer-Country` by default on our stacks.

roborourke:
I did a bunch of testing locally on the header, cookie and query string cache key variation, headers are reflected properly too. Merging in.

PR #73, opened 2019-10-29
Check for REQUEST_URI before using in healthchecks

roborourke:
Checking for REQUEST_URI early in the cloud module breaks in contexts where `$_SERVER['REQUEST_URI']` is not set

### DIFF
--- a/inc/healthcheck/inc/cavalcade/namespace.php
+++ b/inc/healthcheck/inc/cavalcade/namespace.php
@@ -15,7 +15,7 @@ const HEALTHY_THRESHOLD = 900; // 15 mins
  */
 function bootstrap() {
 	add_action( JOB_HOOK, __NAMESPACE__ . '\\set_last_run' );
-	// @codingStandardsIgnoreLine
+	// phpcs:ignore WordPress.WP.CronInterval.ChangeDetected
 	add_filter( 'cron_schedules', __NAMESPACE__ . '\\add_cron_schedule' );
 
 	// Schedule if not already scheduled.
--- a/inc/healthcheck/inc/cavalcade/namespace.php
+++ b/inc/healthcheck/inc/cavalcade/namespace.php
@@ -15,6 +15,7 @@ const HEALTHY_THRESHOLD = 900; // 15 mins
  */
 function bootstrap() {
 	add_action( JOB_HOOK, __NAMESPACE__ . '\\set_last_run' );
+	// @codingStandardsIgnoreLine
 	add_filter( 'cron_schedules', __NAMESPACE__ . '\\add_cron_schedule' );
 
 	// Schedule if not already scheduled.
--- a/inc/performance_optimizations/namespace.php
+++ b/inc/performance_optimizations/namespace.php
@@ -15,11 +15,10 @@ function bootstrap() {
  *
  */
 function increase_set_time_limit_on_async_upload() {
-	if (
-		isset( $_SERVER['REQUEST_URI'] ) &&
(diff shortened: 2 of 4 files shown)


inc/healthcheck/inc/cavalcade/namespace.php
```diff
@@ -15,6 +15,7 @@
  */
 function bootstrap() {
 	add_action( JOB_HOOK, __NAMESPACE__ . '\\set_last_run' );
+	// @codingStandardsIgnoreLine
```
rmccue: !

inc/performance_optimizations/namespace.php
```diff
@@ -17,7 +15,11 @@ function bootstrap() {
  *
  */
 function increase_set_time_limit_on_async_upload() {
-	if ( ini_get( 'max_execution_time' ) < 120 ) {
+	if (
+		isset( $_SERVER['REQUEST_URI'] ) &&
+		strpos( $_SERVER['REQUEST_URI'], '/wp-admin/async-upload.php' ) !== false &&
+		ini_get( 'max_execution_time' ) < 120
+	) {
```
rmccue: Can we flip this and return early to avoid the giganto-if?

inc/healthcheck/inc/cavalcade/namespace.php
roborourke: It was complaining about modifying the cron array. Maybe it's a new sniff

inc/performance_optimizations/namespace.php
roborourke: Not sure if this update is quite what you had in mind. There's a lot of stuff to check either way.

inc/performance_optimizations/namespace.php
```diff
@@ -17,6 +15,9 @@ function bootstrap() {
  *
  */
 function increase_set_time_limit_on_async_upload() {
+	if ( ! isset( $_SERVER['REQUEST_URI'] ) || strpos( $_SERVER['REQUEST_URI'], '/wp-admin/async-upload.php' ) === false ) {
+		return;
+	}
 	if ( ini_get( 'max_execution_time' ) < 120 ) {
```
rmccue: I'd flip this as well; I'd also separate these out into multiple if statements

roborourke: ah I see, so a little family of `if`s

rmccue: 👨‍👩‍👧‍👦

inc/performance_optimizations/namespace.php
```diff
@@ -17,7 +15,14 @@ function bootstrap() {
...
 function increase_set_time_limit_on_async_upload() {
-	if ( ini_get( 'max_execution_time' ) < 120 ) {
-		set_time_limit( 120 );
+	if ( ! isset( $_SERVER['REQUEST_URI'] ) ) {
+		return;
+	}
+	if ( strpos( $_SERVER['REQUEST_URI'], '/wp-admin/async-upload.php' ) === false ) {
+		return;
+	}
+	if ( ini_get( 'max_execution_time' ) >= 120 ) {
```
rmccue: Oh actually, what if the max_execution_time is set to unlimited (`0`)?

roborourke:
```suggest
	$max = ini_get( 'max_execution_time' );
	 if ( $max === 0 || $max >= 120 ) {
```

rmccue:
@roborourke can you do the manual backport here?

roborourke:
@rmccue done & tagged 2.0.3

PR #80, opened 2019-11-11
Enable X-Ray always

joehoyle:
This changes the default for X-Ray to be enabled in all environments, not such ecs and ec2. This is because X-Ray is now used in the Dev Tools too.

See https://github.com/humanmade/altis-dev-tools/issues/23


joehoyle:
It shouldn't cause any issues AFAIK

Issue #77, opened 2019-11-07
Production SQL error on wp-install.php wp_1_options 


tcrsavage:
I opened a Zendesk ticket about this, but think needs looking at by the Altis team specifically.

We saw the issue on staging for a while, but I think Joe also changed the healthcheck url on staging as a temporary fix for a different issue, which was resolved via  https://github.com/humanmade/altis-cloud/pull/70

```
[07-Nov-2019 13:53:13 UTC] WordPress-Datenbank-Fehler Table 'wordpress.wp_1_options' doesn't exist für Abfrage INSERT INTO `wp_1_options` (`option_name`, `option_value`, `autoload`) VALUES ('widget_pages', 'a:1:{s:12:\"_multiwidget\";i:1;}', 'yes') ON DUPLICATE KEY UPDATE `option_name` = VALUES(`option_name`), `option_value` = VALUES(`option_value`), `autoload` = VALUES(`autoload`) # Trace ID: 1-5dc421c8-fc02c9c347d31fdb026a879b von require_once('wp-load.php'), require_once('/srv/www/release-77f902fcd1fe08ff46156d791bf65d3d-1572604843/wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, wp_widgets_init, do_action('widgets_init'), WP_Hook->do_action, WP_Hook->apply_filters, WP_Widget_Factory->_register_widgets, WP_Widget->_register, WP_Widget->get_settings, WP_Widget->save_settings, update_option, add_option, Altis\Cloud\DB->query, wpdb->print_error
```

From Nathanial via Zendesk

> I had a look at the Trace IDs for the PHP errors and they appear to be all related to the Load Balancer healthcheck, which makes the request to `http://{Load Balancer IP}/wp-admin/install.php`. I have a feeling WordPress is not able to correctly attribute this to the default site, and therefore we're seeing it attempt to add the site ID of 1 to the table name. We aren't able to pass a Host header via the healthcheck, so a specific exemption is needed to handle site detection in this case. As we are not experiencing this issue on other stacks, I believe this is confined only to RBMH itself. The Load Balancer does have a specific User Agent, so that can be detected: "ELB-HealthChecker/2.0."

I suspect this can be fixed with a config change, but can't seem to replicate it on local install.

nathanielks:
[...]

PR #81, opened 2019-11-13
Set MULTISITE to prevent $wpdb from setting the database table prefix incorrectly

nathanielks:
Fixes https://github.com/humanmade/altis-cloud/issues/77.

We discovered in some cases the fixes in https://github.com/humanmade/altis-cloud/pull/73 did not adequately prevent errors during healthchecks. Specifically, when `$wpdb` sets the blog prefix, it checks if `MULTISITE` is set and when not follows the standard `{prefix}_{blog id}_table` pattern. This PR sets `MULTISITE` in order to prevent this issue.

Of note is whether or not we should set it to `true` or `false`. We currently detect `WP_INSTALLING` and conditionally set `MULTISITE` if `WP_INSTALLING` is _not_ defined, which makes me think we need to set this to `false` to prevent any weirdness.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -62,6 +62,7 @@ function bootstrap() {
 		isset( $_SERVER['REQUEST_URI'] ) &&
 		$_SERVER['REQUEST_URI'] === '/wp-admin/install.php'
 	) {
+		define( 'MULTISITE', false );
 		define( 'SUBDOMAIN_INSTALL', false );
 	}
 }


nathanielks:
The specific conditional in `$wpdb` is here:

https://github.com/WordPress/WordPress/blob/84c79fcdfdac9fa6ff0b2441ba549dfc100b5ebe/wp-includes/wp-db.php#L984-L988
```php
if ( defined( 'MULTISITE' ) && ( 0 == $blog_id || 1 == $blog_id ) ) {
	return $this->base_prefix;
} else {
	return $this->base_prefix . $blog_id . '_';
}
```

Because `MULTISITE` isn't defined _and_ it sets the `$blog_id` to 1, the first conditional does not get tripped.

PR #85, opened 2019-11-18
Add support for Query Monitor to DB drop-in

### LINKED ISSUE #84: Integrate database drop-in with dev tools
As we have a custom DB class, we should integrate with Query Monitor's extended database logging functionality.

See https://github.com/johnbillion/query-monitor/issues/201 and https://github.com/johnbillion/query-monitor/issues/484

rmccue:
Adds support into our custom DB drop-in for Query Monitor, which provide richer information about database queries, including a full trace and row information. It also allows breaking down queries by component, and rich error data.

Fixes #84.

Before | After
-- | --
Screen Shot 2019-11-18 at 17 35 40 | Screen Shot 2019-11-18 at 17 35 20

Error Before | Error After
-- | --
Screen Shot 2019-11-18 at 17 40 01 | Screen Shot 2019-11-18 at 17 40 09

cc @johnbillion - it'd be neat if the data collection part could be abstracted in QM so this is less brittle if you ever need to change the DB drop-in code.

### DIFF
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -29,14 +29,14 @@ class DB extends LudicrousDB {
 	 *
 	 * @var array
 	 */
-	public $qm_php_vars = array(
+	public $qm_php_vars = [
 		'max_execution_time'  => null,
 		'memory_limit'        => null,
 		'upload_max_filesize' => null,
 		'post_max_size'       => null,
 		'display_errors'      => null,
 		'log_errors'          => null,
-	);
+	];
 
 	/**
 	 * Class constructor
@@ -82,9 +82,9 @@ class DB extends LudicrousDB {
 		}
 
 		$i = count( $this->queries ) - 1;
-		$this->queries[ $i ]['trace'] = new QM_Backtrace( array(
+		$this->queries[ $i ]['trace'] = new QM_Backtrace( [
 			'ignore_frames' => 1,
-		) );
+		] );
 
 		if ( ! isset( $this->queries[ $i ][3] ) ) {
 			$this->queries[ $i ][3] = $this->time_start;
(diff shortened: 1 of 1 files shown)


inc/class-db.php
```diff
@@ -22,9 +24,50 @@ class DB extends LudicrousDB {
...
+		'display_errors'      => null,
+		'log_errors'          => null,
+	];
+
+	/**
+	 * Class constructor
+	 */
+	public function __construct( $args = null ) {
+		foreach ( $this->qm_php_vars as $key => &$val ) {
+			$val = ini_get( $setting );
```
joehoyle: @rmccue WTF is this? This is spitting notices everywhere for me!

rmccue: QM does this in its drop in, so ¯\_(ツ)_/¯ The notices should be silenced though?

joehoyle:
But where is this variable coming from?

-- 

Joe Hoyle

CTO
Human Made

inc/class-db.php
```diff
@@ -33,6 +76,39 @@ function query( $query ) {
...
+					// Please do not report this code as a PHP 7 incompatibility. Observe the surrounding logic.
+					// @codingStandardsIgnoreLine
+					$code = mysql_errno( $this->dbh );
+				}
+			}
+			$this->queries[ $i ]['result'] = new WP_Error( $code, $this->last_error );
+		} else {
+			$this->queries[ $i ]['result'] = $result;
+		}
+
```
johnbillion: I'll see if I can abstract this out into a separate function in QM.

rmccue: Thanks!

PR #89, opened 2019-11-20
Remove unnecessary QM PHP variable gathering

rmccue:
Follow-up for #85.

### DIFF
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -25,32 +25,6 @@ class DB extends LudicrousDB {
 	public $time_spent = 0;
 
 	/**
-	 * Query Monitor PHP variables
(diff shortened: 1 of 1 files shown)


johnbillion:
@rmccue FWIW the original code contained a mistake. In QM's `db.php` the key variable is named `$setting` not `$key`.

This code collects ini config values early on so if any of them change during runtime you'll see a message about it (in the Environment panel). It's not hugely useful so no problem not using it on Altis.

rmccue:
@johnbillion Yeah, I copied the code across verbatim and forgot to clean it up properly; I started renaming the variables, then forgot to finish that or remove it. 🙈

PR #109, opened 2019-12-10
Update batcache to 1.3.1

faishal:


roborourke:
@faishal does this need backporting for v2 & v1? eg. is it a bug fix?

faishal:
@roborourke - yes. it will create different variations based on the ORIGIN header. 

Also, when are we going to release these changes?

roborourke:
@faishal ah so looks like v2 and v1 bundle batcache directly. Not a straightforward backport so will have to PR your upstream change to the `v1-branch` and `v2-branch` directly

roborourke:
@faishal for reference we can tag patch releases any time, there's no set cadence for that. Essentially as soon as something that's a security or bug fix goes in we can tag a patch release.

I've tagged 1.0.4 and 2.0.6 with the batcache update.

PR #113, opened 2019-12-12
Catch the cloudwatch error and display error message

faishal:


inc/audit_log_to_cloudwatch/class-cloudwatch-driver.php
```diff
@@ -125,7 +127,12 @@ public function get_records( $args ) {
 			'queryString'    => $query,
 		];
 
-		$query = cloudwatch_logs_client()->startQuery( $params );
+		try {
+			$query = cloudwatch_logs_client()->startQuery( $params );
```
joehoyle: This should be caught by the caller, and error displayed there. This function should continue to Throw (or return a WP_Error). Calling the function `get_records` should not output an error to the output buffer, as this function could be called in other contexts.

faishal: I looked into stream plugin code and it was not checking WP_Error or catching and throwable exception.
Reference : https://github.com/xwp/stream/blob/e207cf8795542eb73bdce7b80f037ea8e3a22e55/classes/class-db.php#L203

Also, there is no way to filter the list table `no_items` function. 

https://github.com/xwp/stream/blob/e207cf8795542eb73bdce7b80f037ea8e3a22e55/classes/class-list-table.php#L80

joehoyle: Hmm ok- I wonder if we might be better off trigger an error in the error log then. I think the output for this function call still has to be removed. We could still hook in in admin notices or similar to show an error that we store elsewhere in memory

joehoyle: Ping @faishal

faishal: Done, I stored in static variable of the class and moved rendering part on `wp_stream_after_list_table`

inc/audit_log_to_cloudwatch/namespace.php
```diff
@@ -15,9 +16,28 @@ function bootstrap() {
  * This is only done when we we are deemed to have a connection to CloudWatch available.
  *
  * @param string $db_driver
+ *
  * @return string
  */
-function set_stream_db_driver( string $db_driver ) : string {
+function set_stream_db_driver( string $db_driver ): string {
```
roborourke:
```suggestion
function set_stream_db_driver( string $db_driver ) : string {
```

Minor CS fix

PR #119, opened 2020-01-21
Enable redis cache groups

roborourke:
Allows us to use `wp_cache_delete_group` provided by wp-redis.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -219,6 +219,9 @@ function load_object_cache_redis() {
 	if ( ! defined( 'WP_REDIS_DISABLE_FAILBACK_FLUSH' ) ) {
 		define( 'WP_REDIS_DISABLE_FAILBACK_FLUSH', true );
 	}
+	if ( ! defined( 'WP_REDIS_USE_CACHE_GROUPS' ) ) {
+		define( 'WP_REDIS_USE_CACHE_GROUPS', true );
+	}
 
 	Alloptions_Fix\bootstrap();
 	\WP_Predis\add_filters();


joehoyle:
Is there any trade-off here? Otherwise why is this even an option?

roborourke:
@joehoyle as far as I can tell it's just because it diverges from WP's default caching behaviour which is to flatten keys by prepending the group. This enables WP to use redis' built in structured storage.

The trade off is that this change will invalidate the existing caches. The benefit is you then get to use `wp_cache_delete_group()`, so you could use an SQL query hash as the key for instance then delete all those cached queries when an item is added/deleted/updated. Current use case is Cavalcade and the updates here https://github.com/humanmade/Cavalcade/pull/91

roborourke:
Reference: https://pantheon.io/blog/wp-redis-v030-cache-groups

rmccue:
> You will also need to manage natural cache expiration via group-wide time-to-live (TTL) settings; when enabled, WP Redis doesn't support setting expiration on individual keys.

That seems like a pretty hefty limitation I think as it diverges from WP's behaviour in a significant way.

roborourke:
@rmccue ah interesting, in that case we could explicitly use `$object_cache->add_redis_hash_groups()` instead. Allows to use structured storage for a specific group instead. In which case this would just mean we need to document it or make it a configurable thing in `composer.json`. In the case of plugins they'll have to check for both the group delete function and add the relevant hash groups using the above method themselves.

joehoyle:
Could we include the cachettl in the group name? For many (most) things that don't specify a TTL can all use the same group.

Though I guess `wp_cache_delete_group` would then also need to take a TTL.

rmccue:
All we really care about here is being able to delete a whole group; can we instead do a [KEYS](https://redis.io/commands/keys) and then a [DEL](https://redis.io/commands/del)?

```
KEYS "{salt}{prefix}{group}:*"
DEL {result}
```

(For performance, `SCAN 0 MATCH "{salt}{prefix}{group}:*" may be better than KEYS, but conceptually I mean.)

rmccue:
Of note: https://stackoverflow.com/questions/27976255/redis-wildcard-delete-script-using-eval-scan-and-del-returns-write-commands-n

roborourke:
Oh tidy, I was wondering if wildcard deletion was supported. So we can write our own `wp_cache_delete_group()` function, albeit with a slightly different name. Alternatively we could try contributing your method upstream as a fallback.

PR #120, opened 2020-01-22
R	emove extra data in xray requests

joehoyle:
Current we send way too much data in `$_SERVER` in X-Ray requests, which is also potentially exposing data that we might not want to.


rmccue:
Backporting to v3 only, requires X-Ray plugin 1.2+.

Issue #124, opened 2020-01-29
Add environment indicator to admin bar


roborourke:
Moved here from https://github.com/humanmade/altis-cms/issues/113

Acceptance criteria:

- [x] add a solution to differentiate environments via colour / other indicator
  - add the feature to a new namespace directory in the `inc` directory
  - evaluate https://github.com/tomjn/Local-Indicator for use, might be easier to use the approach from that and add the functionality directly here using `Altis\get_environment_type()` for example
- [x] indicator should show environment type in the top level menu item
  - indicator sub menu should show:
     - [x] the stack name
     - [x] a link to the stack on Altis dashboard (if env is not === local)
- [x] indicator should only show for admins or those with an `environment_indicator` capability

@Camila-villegas can you provide some guidance on the colours we could use to indicate the following environment types?

- local (safe to make any changes)
- development (unstable, safe to make changes and test things out)
- staging (semi-stable, used for client Quality Assurance mostly)
- production (exercise caution when making changes / edits etc)

TODO:
- [ ] Implement CSS when decisions on design are made
- [ ] Test Altis dashboard URL is correct for variety of environments (not sure how to do that on local - added by @dashaluna )

roborourke:
Noted by @dashaluna the Local Indicator plugin looks like this by default:

![Local Indicator](https://user-images.githubusercontent.com/2185972/73377709-5cb47480-42b7-11ea-98f4-bc79fcb8edc9.png)

There are other design opportunities to consider here:

- could use a coloured border above the admin bar
- icon with a different colour to indicate environment type
- text makes it clearer what it means

dashaluna:
@roborourke you ok closing this issues as PR has been merged? Don't think anything else is left.

roborourke:
@dashaluna yep, nicely done here. For future PRs you can add the text “Fixes #xyz” to the PR comment and it’ll automatically close it when the PR is merged.

PR #126, opened 2020-01-30
Add menu and submenu to display ENV details

dashaluna:
See #124

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
 		"files": [
 			"inc/healthcheck/namespace.php",
 			"inc/healthcheck/cavalcade/namespace.php",
+			"inc/environment_indicator/namespace.php",
 			"inc/namespace.php",
 			"wp-config.php"
 		]
--- a/inc/environment_indicator/namespace.php
+++ b/inc/environment_indicator/namespace.php
@@ -10,12 +10,12 @@ function bootstrap() {
 	add_action( 'admin_bar_init', __NAMESPACE__ . '\\enqueue_admin_scripts' );
 	// Allow other modules to add sub-menu items to Altis admin bar menu first.
 	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_admin_bar_env_info', 15 );
+	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_admin_bar_dashboard_link', 15 );
 }
 
 /**
  * Add menu item to admin bar to display environment info:
  * - environment type
- * - environment name
  * - URL to Altis dashboard for the environment
  *
  * @param WP_Admin_Bar $wp_admin_bar WP_Admin_Bar instance, passed by reference.
@@ -48,22 +48,40 @@ function add_admin_bar_env_info( WP_Admin_Bar $wp_admin_bar ) {
 
 	// Add environment indicator to the Altis logo menu item in the admin bar.
 	$node = $wp_admin_bar->get_node( 'altis' );
-	$node->meta = [
-		'html' => sprintf( '<span class="altis-env-indicator">%s</span>', $envs[ get_environment_type() ] ),
-	];
+	$node->title .= ' ' . sprintf( '<span class="altis-env-indicator">%s</span>', $envs[ get_environment_type() ] );
 	$wp_admin_bar->add_menu( $node );
+}
+
+/**
+ * Add menu item to admin bar to display environment info:
+ * - URL to Altis dashboard for the environment
+ *
+ * @param WP_Admin_Bar $wp_admin_bar WP_Admin_Bar instance, passed by reference.
+ */
+function add_admin_bar_dashboard_link( WP_Admin_Bar $wp_admin_bar ) {
+	if ( ! is_admin_bar_showing() ) {
+		return;
+	}
+
+	if ( ! current_user_can( 'manage_options' ) ) {
+		return;
+	}
 
 	// Stop - no Altis dashboard URL is available. No need to add sub-menu item.
-	if ( ! defined( 'HM_ENV_REGION' ) || get_environment_type() === 'local' ) {
-		return;
+	if ( ! defined( 'HM_ENV_REGION' ) ) {
+		$title = __( 'Open Altis Dashboard', 'altis' );
+		$url = 'https://dashboard.altis-dxp.com/';
+	} else {
+		$title = __( 'Open in Altis Dashboard', 'altis' );
+		$url = esc_url( sprintf( 'https://dashboard.altis-dxp.com/#/%s/%s', HM_ENV_REGION, get_environment_name() ) );
 	}
 
 	// Add environment's Altis dashboard URL as a sub-menu item to Altis logo menu in the admin bar.
 	$wp_admin_bar->add_menu( [
 		'parent' => 'altis',
 		'id'     => 'altis-env-stack-url',
-		'title'  => __( 'Open in Altis Dashboard', 'altis' ),
-		'href'   => esc_url( sprintf( 'https://dashboard.altis-dxp.com/#/%s/%s', HM_ENV_REGION, get_environment_name() ) ),
+		'title'  => $title . ' <span class="dashicons-before dashicons-external"></span>',
+		'href'   => $url,
 		'meta'   => [
 			'target' => '_blank',
 		],
--- a/inc/environment_indicator/namespace.php
+++ b/inc/environment_indicator/namespace.php
@@ -8,7 +8,8 @@ use WP_Admin_Bar;
 
 function bootstrap() {
 	add_action( 'admin_bar_init', __NAMESPACE__ . '\\enqueue_admin_scripts' );
-	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_admin_bar_env_info' );
+	// Allow other modules to add sub-menu items to Altis admin bar menu first.
+	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_admin_bar_env_info', 15 );
 }
 
 /**
@@ -30,14 +31,6 @@ function add_admin_bar_env_info( WP_Admin_Bar $wp_admin_bar ) {
 		return;
 	}
 
-	// Add environment menu item to the admin bar.
-	$wp_admin_bar->add_menu( [
-		'id'   => 'altis-env-indicator',
-		'meta' => [
-			'class' => 'altis-env-indicator--' . get_environment_type(),
-		],
-	] );
-
 	// Environment types text to be displayed in the admin bar.
 	$envs = [
 		'local'       => esc_html_x( 'local', 'Server environment type', 'altis' ),
@@ -52,23 +45,23 @@ function add_admin_bar_env_info( WP_Admin_Bar $wp_admin_bar ) {
 		),
 	];
 
-	// Environment type sub-menu item.
-	$wp_admin_bar->add_menu( [
-		'id'     => 'altis-env-stack-type',
-		'title'  => $envs[ get_environment_type() ],
-		'parent' => 'altis-env-indicator',
-	] );
+	// Add environment indicator to the Altis logo menu item in the admin bar.
+	$node = $wp_admin_bar->get_node( 'altis' );
+	$node->meta = [
+		'html' => sprintf( '<span class="altis-env-indicator">%s</span>', $envs[ get_environment_type() ] ),
+	];
+	$wp_admin_bar->add_menu( $node );
 
-	// Stop - no Altis dashboard URL is available.
+	// Stop - no Altis dashboard URL is available. No need to add sub-menu item.
 	if ( ! defined( 'HM_ENV_REGION' ) || get_environment_type() === 'local' ) {
 		return;
 	}
 
-	// Environment's Altis dashboard URL sub-menu item.
+	// Add environment's Altis dashboard URL as a sub-menu item to Altis logo menu in the admin bar.
 	$wp_admin_bar->add_menu( [
+		'parent' => 'altis',
 		'id'     => 'altis-env-stack-url',
 		'title'  => __( 'Open in Altis Dashboard', 'altis' ),
-		'parent' => 'altis-env-indicator',
 		'href'   => esc_url( sprintf( 'https://dashboard.altis-dxp.com/#/%s/%s', HM_ENV_REGION, get_environment_name() ) ),
 		'meta'   => [
 			'target' => '_blank',
--- a/inc/environment_indicator/namespace.php
+++ b/inc/environment_indicator/namespace.php
@@ -7,6 +7,7 @@ use function Altis\get_environment_type;
 
 function bootstrap() {
 	add_action( 'wp_before_admin_bar_render', __NAMESPACE__ . '\\add_admin_bar_env_info' );
+	add_action( 'admin_bar_init', __NAMESPACE__ . '\\enqueue_admin_scripts' );
 }
 
 /**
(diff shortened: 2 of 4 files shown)


dashaluna:
@rob here is the initial work. Let me know if it's on the right track.
Also, I'm a bit unsure how to display `a link to the stack on Altis dashboard` is there a function?

Screenshot 2020-01-30 15 46 12

roborourke:
@dashaluna sorry, I was thinking this would go in the admin bar rather the main admin menu so it's visible whenever you're logged in.

Regarding linking to Vantage there isn't a function but you can find the URL for a stack at https://dashboard.altis-dxp.com/ and build it dynamically in PHP eg. `https://dashboard.altis-dxp.com/#/<region>/<stack-name>`

Region you can get from `HM_ENV_REGION` and the stack name is `Altis\get_environment_name()`.

roborourke:
I like the icon you've chosen btw 👍

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,65 @@
...
+
+use function Altis\get_environment_name;
+use function Altis\get_environment_type;
+
+function bootstrap() {
+	add_action( 'admin_menu', __NAMESPACE__ . '\\add_menu_page' );
+}
+
+function add_menu_page() {
+	\add_menu_page(
```
roborourke: I was after adding this to the admin bar rather than admin menu, sorry it wasn't clearer

dashaluna:
Thanks @roborourke for the feedback.
How about this controversial approach: to add ENV info to both sidebar (where it is now) and to the admin bar.

My thinking is:
- admin bar could be switched off for a user even if they are admin, so the info in the main menu always gonna be there
- the more reminders the better - they are only little menu items.

roborourke:
That’s true the admin bar can switched off but it’s rare. I think given the number of menu items some sites have it’d just be lost, plus the page itself doesn’t really show enough data to warrant a whole page I think. Let’s stick with admin bar only for now

composer.json
```diff
@@ -13,6 +13,7 @@
 		"files": [
 			"inc/healthcheck/namespace.php",
 			"inc/healthcheck/cavalcade/namespace.php",
+			"inc/environment_indicator/namespace.php",
```
dashaluna: @roborourke I moved loading into composer and run `composer dump-autoload` and the changes don't appear.
I've also tried to remove symlink to the package and re-install it again, still no bingo :(

What am I missing here?

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,79 @@
...
+	$capabilities[] = 'manage_options';
+
+	if ( is_multisite() ) {
+		$capabilities[] = 'manage_network_options';
+	}
+
+	$capabilities = apply_filters( 'altis_env_indicator_capabilities', $capabilities );
+	foreach ( $capabilities as $cap ) {
+		if ( ! current_user_can( $cap ) ) {
+			return;
```
roborourke: Is the intention that the user should have all of the capabilities listed in `altis_env_indicator_capabilities` to be able to see it? This will stop executing the function as soon as a user lacks even one of the specified capabilities which I'm not sure is what you're going for.

I think it might be better to approach this the other way around. Check for a capability called `view_environment_indicator` or `is_super_admin()`.

composer.json
roborourke: You appear to have removed the bit that called `Environment_Indicator\bootstrap()` in the main `inc/namespace.php` file.

dashaluna: Classic :) thanks!

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,79 @@
...
+		$altis_dashboard_url = sprintf( 'https://dashboard.altis-dxp.com/#/%s/%s', esc_url( HM_ENV_REGION ), esc_url( get_environment_name() ) );
+	}
+
+	$wp_admin_bar->add_menu( [
+		'id'     => 'altis-env-stack-url',
+		'title'  => sprintf( __( 'Altis dashboard: %s', 'altis' ), esc_html( get_environment_name() ) ),
+		'parent' => 'altis-env-indicator',
+		'href'   => $altis_dashboard_url,
+		'meta'   => [
+			'target' => '__blank',
```
roborourke: Single underscore for target value

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,79 @@
...
+#wpadminbar #wp-admin-bar-altis-env-indicator>.ab-item:focus,
+#wpadminbar #wp-admin-bar-altis-env-indicator>.ab-item:visited {
+	background: ' . $indicator_colour . ' ! important;
+}
+</style>';
+	$bar_item_style = apply_filters( 'altis_env_indicator_style', $bar_item_style, get_environment_type() );
+
+	// Add menu items to admin bar.
+	$wp_admin_bar->add_menu( [
+		'id'    => 'altis-env-indicator',
```
roborourke: Let's add a class name too to indicate the environment type eg. `altis-env-indicator--<type>`

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,79 @@
...
+
+	$bar_item_style = '<style>
+#wpadminbar #wp-admin-bar-altis-env-indicator>.ab-item:before {
+	content: "\f547";
+}
+#wpadminbar #wp-admin-bar-altis-env-indicator,
+#wpadminbar #wp-admin-bar-altis-env-indicator>.ab-item:hover,
+#wpadminbar #wp-admin-bar-altis-env-indicator>.ab-item:focus,
+#wpadminbar #wp-admin-bar-altis-env-indicator>.ab-item:visited {
+	background: ' . $indicator_colour . ' ! important;
```
roborourke: I think we should control the indicator colours ourselves in a CSS file, we have an assets folder in the root of this package. See the branding namespace for an example.

Using the different class names we can set the colour. I'd like to avoid using `!important` if possible too.

inc/environment_indicator/namespace.php
dashaluna: Hm, first of all I couldn't find the `view_environment_indicator` capability in the code. Where is it defined?
I would argue that we should display it for all `admins` not just `super_admins` as `admins` still can make changes with side effects they better be sure and notified what env they are on. So, should we keep it to `manage_options`?

dashaluna:
This is how it looks so far:

Screenshot 2020-01-31 15 22 59

I added a todo list to the main issue to note that CSS needs implementing when designs are ready.

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,79 @@
...
+		'title' => $bar_item_style . $indicator_text,
+	] );
+
+	// Construct Altis dashboard URL for the environment.
+	$altis_dashboard_url = '';
+	if ( defined( 'HM_ENV_REGION' ) && get_environment_type() !== 'local' ) {
+		$altis_dashboard_url = sprintf( 'https://dashboard.altis-dxp.com/#/%s/%s', esc_url( HM_ENV_REGION ), esc_url( get_environment_name() ) );
+	}
+
+	$wp_admin_bar->add_menu( [
```
roborourke: We shouldn't show this link at all if there's no URL. The link text won't make sense.

inc/environment_indicator/namespace.php
roborourke: Colours based on our branding guidelines https://drive.google.com/drive/folders/1GnnbPJ87pnnjbklts656byxkbfutr1J6:

- local: `#5AC5B1` Green
- development: `#06A9CE` Aqua
- staging: `#F8DA6D` Yellow
- production: `#ED7B9D` Pink

Make sure the link text is visible, the dark blue `#152A4E` should be suitable.

inc/environment_indicator/namespace.php
roborourke: This link text is likely to be very long so let's break it up. Have two child nodes, the first one should show the stack name, the second should be a link with the text "Open in Altis Dashboard"

inc/environment_indicator/namespace.php
roborourke: A capability is just a string passed to `current_user_can()`, by doing that you've essentially created a capability that developer can assign to different roles. There's nothing more to do.

Good point re. single site admins. To be honest I'm not totally sure - a super admin is more likely to have access to staging and development sites and will have the context for what it means to be on production, whereas a single site admin is more likely to be working on production by default anyway. The goal is really to hide this from people who are most likely to just be doing their day to day work.

My thinking re. super admins is because all Altis sites are multisite by default. How about we default to super admins and make it a configuration option to show the indicator regular site admins too.

dashaluna:
@roborourke thanks for the feedback. Here is how it looks so far. Ignore "unknown" for the stack name, I just spoofed the environment types to see the colours.

Screenshot 2020-02-03 11 49 10
Screenshot 2020-02-03 11 49 34
Screenshot 2020-02-03 11 49 50
Screenshot 2020-02-03 11 50 07

Questions:
1. Should we keep the info the last item on the admin bar? I can make it the first one instead, like so:
Screenshot 2020-02-03 11 44 38
2. Any other styling adjustments?

roborourke:
@dashaluna colours are nice I think - I may have been wrong about the text colour though looking at it.

It's an odd one really to try and make it pretty. Because it's a big block of colour we could perhaps force it all the way over the right so it's the last item in the admin bar.

Alternatively we could make the colour into an indicator rather than the whole background. Like a circle following the environment name. To really compact things we could also move the environment name into the submenu. People would learn what the colours indicate then.

Can we try removing the background colour, have the computer icon in white followed by a circle of the indicator colour? Eg:

💻🔴

roborourke:
Another reason for using the colour as an indicator - our branding guidelines actually state that's specifically what they're for - the higher contrast colours are more accessible to use as backgrounds.

inc/environment_indicator/namespace.php
```diff
@@ -27,45 +28,42 @@ function add_admin_bar_env_info() {
...
+			'id'     => 'altis-env-stack-name',
+			'title'  => esc_html( get_environment_name() ),
+			'parent' => 'altis-env-indicator',
+		] );
+
+		$wp_admin_bar->add_menu( [
+			'id'     => 'altis-env-stack-url',
+			'title'  => __( 'Open in Altis Dashboard', 'altis' ),
+			'parent' => 'altis-env-indicator',
+			'href'   => sprintf( 'https://dashboard.altis-dxp.com/#/%s/%s', esc_url( HM_ENV_REGION ), esc_url( get_environment_name() ) ),
```
roborourke: Just realised `esc_url()` is used incorrectly here - use it to escape the full URL, not individual pieces of it

dashaluna:
@roborourke below are some examples. 

Screenshot 2020-02-03 12 35 56
Screenshot 2020-02-03 12 36 09
Screenshot 2020-02-03 12 36 27
Screenshot 2020-02-03 12 36 38
Screenshot 2020-02-03 13 04 26
Screenshot 2020-02-03 13 05 55
Screenshot 2020-02-03 13 06 30
Screenshot 2020-02-03 13 06 52

I personally prefer a single computer icon with a solid bg colour. I also noticed that query monitor displays a red bg when there are errors. In out case the menu item is small and colours are not as bright, I think that looks the best. Also having a square bg like that allows to differentiate between green and blue better as with smaller footage they look very similar.
Screenshot 2020-02-03 13 13 19
Screenshot 2020-02-03 13 13 37
Screenshot 2020-02-03 13 15 30
Screenshot 2020-02-03 13 15 46

What do you think?

roborourke:
Thanks @dashaluna! I think I like the dot one best personally. My last idea here is that we could use a different shape too to make it a bit more accessible eg.

- local: circle
- development: square
- staging: diamond
- production: triangle

There are examples for making the shapes in CSS here but lets get a bit feedback first. This is looking really cool though. https://css-tricks.com/the-shapes-of-css/

roborourke:
Ok, after a bit more back and forth I think it's going to be best to show the text as it's less obscure than colours & icons:

Screenshot 2020-02-03 14 19 29

For development shorten the text to `dev` and for production shorten it to `prod` but wrap in an `abbr` tag eg:

```html
<abbr title="production">prod</abbr>
```

Use the Aqua colour #06A9CE as the background.

To extend the current logo menu item do:

```php
$node = $wp_admin_bar->get_node( 'altis );
$node->meta = [ 'html' =>  '<span class="altis-env-indicator">staging</span>' ];
$wp_admin_bar->add_menu( $node );
```

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,76 @@
...
+	$env_type_text = apply_filters( 'altis_env_indicator_text', $env_type_text, get_environment_type() );
+
+	$wp_admin_bar->add_menu( [
+		'id'     => 'altis-env-stack-type',
+		'title'  => $env_type_text,
+		'parent' => 'altis-env-indicator',
+	] );
+
+	// Display environment info for non-local stacks.
+	if ( defined( 'HM_ENV_REGION' ) && get_environment_type() !== 'local' ) {
```
roborourke: It's often neater to use the "return early" pattern, so rather than wrapping & indenting large pieces of code in an `if` statement reverse the logic and `return` inside the `if` statement.
```suggestion
	if ( ! defined( 'HM_ENV_REGION' ) || get_environment_type() === 'local' ) {
		return;
	}
```

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,76 @@
...
+		$wp_admin_bar->add_menu( [
+			'id'     => 'altis-env-stack-name',
+			'title'  => esc_html( get_environment_name() ),
+			'parent' => 'altis-env-indicator',
+		] );
+
+		// Environment's Altis dashboard URL sub-menu item.
+		$wp_admin_bar->add_menu( [
+			'id'     => 'altis-env-stack-url',
+			'title'  => __( 'Open in Altis Dashboard', 'altis' ),
```
roborourke: Let's add the external link icon to this too: https://developer.wordpress.org/resource/dashicons/#external

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,76 @@
...
+	$wp_admin_bar->add_menu( [
+		'id'     => 'altis-env-stack-type',
+		'title'  => $env_type_text,
+		'parent' => 'altis-env-indicator',
+	] );
+
+	// Display environment info for non-local stacks.
+	if ( defined( 'HM_ENV_REGION' ) && get_environment_type() !== 'local' ) {
+		// Environment name sub-menu item.
+		$wp_admin_bar->add_menu( [
```
roborourke: Realised for some clients this will be reallllllllllllllly long so lets drop it in favour of just the Altis Dashboard link.

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,76 @@
...
+	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_admin_bar_env_info' );
+}
+
+/**
+ * Add menu item to admin bar to display environment info:
+ * - environment type
+ * - environment name
+ * - URL to Altis dashboard for the environment
+ */
+function add_admin_bar_env_info() {
```
roborourke: The `admin_bar_menu` action passes the admin bar class through so you don't need the `global $wp_admin_bar;` in there eg:
 ```suggestion
function add_admin_bar_env_info( WP_Admin_Bar $wp_admin_bar ) {
```

You'll need to `use` the `WP_Admin_Bar` class after the namespace declaration.

File: inc/environment_indicator/namespace.php [hunk 13]
```diff
@@ -0,0 +1,76 @@
...
+	$wp_admin_bar->add_menu( [
+		'id'   => 'altis-env-indicator',
+		'meta' => [
+			'class' => 'altis-env-indicator--' . get_environment_type(),
+		],
+	] );
+
+	// Environment type sub-menu item.
+	$env_type_text = strtoupper( get_environment_type() );
+	$env_type_text = apply_filters( 'altis_env_indicator_text', $env_type_text, get_environment_type() );
```
roborourke: I'm not sure this filter is really necessary. I think instead we could have an array of translatable strings eg:
```php
$envs = [
    'local' => _x( 'local', 'Server environment type', 'altis' ),
    'development' => _x( 'dev', 'Server environment type', 'altis' ),
    'staging' => _x( 'staging', 'Server environment type', 'altis' ),
    'production' => _x( 'prod', 'Server environment type', 'altis' ),
];
```

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,76 @@
...
+ * - URL to Altis dashboard for the environment
+ */
+function add_admin_bar_env_info() {
+	if ( ! is_admin_bar_showing() ) {
+		return;
+	}
+
+	// Specify capabilities for which environment indicator is shown in the admin bar.
+	$capability = apply_filters( 'altis_env_indicator_capability', 'manage_options' );
+	if ( ! current_user_can( $capability ) ) {
```
roborourke: Change of approach again here - this is really about differentiating the environments, at the same time we don't really want to affect production as it won't be relevant to most people using that.

Let's make it so that it always shows on `local`, `development` and `staging` by default, then filter that value with a hook called `altis.show_environment_indicator`. That way we can let our engineers decide on any complex logic need or want for this and we can consider incorporating it down the line.

```php
$show_indicator = in_array( get_environment_type(), [ 'local', 'development', 'staging' ], true );
$show_indicator = apply_filters( 'altis.show_environment_indicator', $show_indicator );
if ( ! $show_indicator ) {
     return;
}
```

rmccue: Could we instead allow returning false from `altis_env_indicator_text` and having that skip render?

roborourke: This is clearer IMO rather than overloading another filter and giving it multiple uses. I can't really think of another reason to filter the text that might not change the purpose of the feature.

inc/environment_indicator/namespace.php
dashaluna: I've done it via CSS but somehow I'm seeing a empty square instead of the icon. Not sure what I'm missing. Interestingly enough.. if I use `:before` then it shows ok, but not with `:after` 🤔

dashaluna:
Here is how it looks:
Screenshot 2020-02-04 13 49 28
Screenshot 2020-02-04 13 50 05

Added a comment here https://github.com/humanmade/altis-cloud/pull/126#discussion_r374679398 re the square box instead of a dashicon icon.

inc/environment_indicator/namespace.php
```diff
@@ -0,0 +1,78 @@
...
+		'development' => sprintf( '<abbr title="%s">%s</abbr>',
+			esc_attr_x( 'development', 'Server environment type - full form', 'altis' ),
+			esc_html_x( 'dev', 'Server environment type - abbreviated', 'altis' )
+		),
+		'staging'     => esc_html_x( 'staging', 'Server environment type', 'altis' ),
+		'production'  => sprintf( '<abbr title="%s">%s</abbr>',
+			esc_attr_x( 'production', 'Server environment type - full form', 'altis' ),
+			esc_html_x( 'prod', 'Server environment type - abbreviated', 'altis' )
+		),
+	];
```
dashaluna: @roborourke I went ahead and used the `esc` functions here as it's most straightforward.
However, I know we should escape late.. should I instead do this on line https://github.com/humanmade/altis-cloud/pull/126/files#diff-dbc927f3487c6469f372e472b9e3cbdcR52 and use `whitelist_html()` to list allowed HTML?

inc/environment_indicator/namespace.php
roborourke: @dashaluna maybe need to set the font family to dashicons. Let me check

roborourke: oh, use an empty span after the text for the icon instead so `:before` can be used

inc/environment_indicator/namespace.php
roborourke: I think this is fine tbh, nicely done

roborourke:
Made a last tweak to fix the external link icon. Thanks for pushing this through and doing all the iterations @dashaluna, really good work.

dashaluna:
@roborourke thanks for letting me know. And making the final touches.

PR #130, opened 2020-02-05
Fix Twenty Twenty frontend display issue

### LINKED ISSUE #129: Visual bug with environment text on frontend
Screen Shot 2020-02-05 at 15 04 14

From #124 and #126.

This is happening in Firefox (73b); may be due to Firefox's different handling of flex-basis.

roborourke:
Fixes #129

### DIFF
--- a/assets/environment-indicator.css
+++ b/assets/environment-indicator.css
@@ -13,10 +13,8 @@
 	margin-top: 6px;
 	padding: 2px 4px;
 	border-radius: 4px;
-}
-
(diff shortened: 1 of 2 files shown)


rmccue:
Looks like this is actually from Twenty Twenty, which is setting `* { word-break: break-word; }`. Setting `word-break: normal` on our element fixes it and might be a better fix here?

roborourke:
Yeah agreed, done 👍

PR #133, opened 2020-02-06
Log remote requests made by the AWS SDK

### LINKED ISSUE #125: Log remote requests made by the AWS SDK
I'm seeing some huge perf issues where being able to dig into which AWS requests are made would be super useful. Because AWS SDK requests go via Guzzle rather than via WP_Http they're not currently logged.

(I know @johnbillion ran into similar things with Query Monitor's HTTP logging, so maybe there's some solution he saw there?)

dashaluna:
Fixes #125

### DIFF
--- a/load.php
+++ b/load.php
@@ -40,3 +40,11 @@ add_action( 'altis.modules.init', function () {
 
 	register_module( 'cloud', __DIR__, 'Cloud', $default_settings, __NAMESPACE__ . '\\bootstrap' );
 } );
+
+// Early hook for logging AWS SDK HTTP requests.
+add_filter( 'altis.aws_sdk.params', function ( $params ) {
+	if ( ! function_exists( 'HM\\Platform\\XRay\\on_hm_platform_aws_sdk_params' ) ) {
+		return $params;
+	}
+	return on_hm_platform_aws_sdk_params( $params );
+} );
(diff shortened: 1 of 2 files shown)


roborourke:
Tested this out locally with the `audit-log-to-cloudwatch` feature on and the SDK params filter is applied.

@joehoyle @rmccue can you think of any other key things to help test this locally? Can you see any issue with where the filter is?

rmccue:
> @joehoyle @rmccue can you think of any other key things to help test this locally? Can you see any issue with where the filter is?

Let's roll it onto a development stack and test against the live service.

roborourke:
Will update `platform-test` to run this branch 👍

load.php
```diff
@@ -39,3 +40,6 @@
 
 	register_module( 'cloud', __DIR__, 'Cloud', $default_settings, __NAMESPACE__ . '\\bootstrap' );
 } );
+
+// Early hook for logging AWS SDK HTTP requests.
+add_filter( 'altis.aws_sdk.params', __NAMESPACE__ . '\\add_aws_sdk_xray_callback' );
```
rmccue: We need to separately hook into S3 Upload's copy of the SDK currently.

```suggestion
add_filter( 'altis.aws_sdk.params', __NAMESPACE__ . '\\add_aws_sdk_xray_callback' );
add_filter( 's3_uploads_s3_client_params', __NAMESPACE__ . '\\add_aws_sdk_xray_callback' );
```

rmccue:
With the above change, I'm now seeing requests come in, however the data on them appears to be wrong. See Slack for more.

rmccue:
Filed https://github.com/humanmade/aws-xray/issues/54 for the invalid-ish data.

We can move ahead with this in the meantime. With the above change, it should be good to go.

rmccue:
Ah, we probably need to add `aws_ses_wp_mail_ses_client_params` too.

roborourke:
Woot

joehoyle:
Do we need to do anything else for _this_ PR, or shall we merge and add follow-ups for the other items?

roborourke:
Added AWS SES params filter as well.

PR #134, opened 2020-02-10
Add link to open support ticket from Altis menu

### LINKED ISSUE #132: Add link to open support ticket from Altis menu
Alongside #126, we should also link directly to the support ticket page.

The format for the link should be `https://dashboard.altis-dxp.com/#/support/new?applications[]={stack-name}` if we have the environment name, or `https://dashboard.altis-dxp.com/#/support/new` otherwise.

dashaluna:
Fixes #132 

@roborourke let me know if it's ok to keep in within the env indicator code? Otherwise, I'll move it into a separate folder (let me know the name of it).

### DIFF
--- a/inc/environment_indicator/namespace.php
+++ b/inc/environment_indicator/namespace.php
@@ -6,6 +6,8 @@ use function Altis\get_environment_name;
 use function Altis\get_environment_type;
 use WP_Admin_Bar;
 
+const SUPPORT_URL = 'https://dashboard.altis-dxp.com/#/support/new';
+
 function bootstrap() {
 	add_action( 'admin_bar_init', __NAMESPACE__ . '\\enqueue_admin_scripts' );
 	// Allow other modules to add sub-menu items to Altis admin bar menu first.
@@ -100,15 +102,17 @@ function add_admin_bar_support_ticket_link( WP_Admin_Bar $wp_admin_bar ) {
 	}
 
 	$env_name    = get_environment_name();
-	$support_url = 'https://dashboard.altis-dxp.com/#/support/new';
-	$support_url = 'unknown' === $env_name ? $support_url : "{$support_url}?applications[]={$env_name}";
+	$support_url = SUPPORT_URL;
+	if ( 'unknown' !== $env_name ) {
+		$support_url .= sprintf( '?applications[]=%s', urlencode( $env_name ) );
+	}
 
 	// Add support ticket URL as a sub-menu item to the Altis logo menu in the admin bar.
 	$wp_admin_bar->add_menu( [
 		'parent' => 'altis',
 		'id'     => 'altis-support-ticket',
 		'title'  => __( 'Open Support Ticket', 'altis' ) . ' <span class="dashicons-before dashicons-external"></span>',
-		'href'   => esc_url( $support_url ),
+		'href'   => $support_url,
 		'meta'   => [
 			'target' => '_blank',
 		],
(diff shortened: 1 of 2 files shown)


inc/environment_indicator/namespace.php
```diff
@@ -88,6 +89,31 @@ function add_admin_bar_dashboard_link( WP_Admin_Bar $wp_admin_bar ) {
...
+ *
+ * @param WP_Admin_Bar $wp_admin_bar WP_Admin_Bar instance, passed by reference.
+ */
+function add_admin_bar_support_ticket_link( WP_Admin_Bar $wp_admin_bar ) {
+	if ( ! is_admin_bar_showing() ) {
+		return;
+	}
+
+	$env_name    = get_environment_name();
+	$support_url = 'unknown' !== $env_name ? "https://dashboard.altis-dxp.com/#/support/new?applications[]={$env_name}" : 'https://dashboard.altis-dxp.com/#/support/new';
```
rmccue: This line is super long; can we move the base URL (`.../support/new`) to a constant?

Also, this should use either `sprintf` or `add_query_arg` for the argument, and it needs to be URL encoded.

dashaluna: @rmccue thanks.

`add_query_arg` adds the args before the `#` in the URL, so getting `https://dashboard.altis-dxp.com/?applications%5B%5D=test#/support/new`, I guess anchor comes after the args in the URL structure.

As for `sprintf` I remember a feedback that URL should be escaped in full, not parts. I've updated code. Let me know if that works.

inc/environment_indicator/namespace.php
```diff
@@ -88,6 +89,32 @@ function add_admin_bar_dashboard_link( WP_Admin_Bar $wp_admin_bar ) {
...
+ * @param WP_Admin_Bar $wp_admin_bar WP_Admin_Bar instance, passed by reference.
+ */
+function add_admin_bar_support_ticket_link( WP_Admin_Bar $wp_admin_bar ) {
+	if ( ! is_admin_bar_showing() ) {
+		return;
+	}
+
+	$env_name    = get_environment_name();
+	$support_url = 'https://dashboard.altis-dxp.com/#/support/new';
+	$support_url = 'unknown' === $env_name ? $support_url : "{$support_url}?applications[]={$env_name}";
```
roborourke: Ternary operators are a bit harder to read in general. In this instance you're not actually modifying the URL when the condition is true so a basic `if` statement would be a better fit and easier to understand:
```suggestion
	if ( $env_name !== 'unknown' ) {
		$support_url .= sprintf( '?applications[]=%s', urlencode( $env_name ) );
	}
```

inc/environment_indicator/namespace.php
```diff
@@ -88,6 +89,32 @@ function add_admin_bar_dashboard_link( WP_Admin_Bar $wp_admin_bar ) {
...
+ *
+ * @param WP_Admin_Bar $wp_admin_bar WP_Admin_Bar instance, passed by reference.
+ */
+function add_admin_bar_support_ticket_link( WP_Admin_Bar $wp_admin_bar ) {
+	if ( ! is_admin_bar_showing() ) {
+		return;
+	}
+
+	$env_name    = get_environment_name();
+	$support_url = 'https://dashboard.altis-dxp.com/#/support/new';
```
rmccue: This should be a `const` in the namespace rather than a variable here

roborourke: Think it's helpful to explain why you'd do that as well

dashaluna:
Thank you both for feedback and patience(!)
Hopefully refactored code is ok :)

PR #140, opened 2020-02-18
Upgrade batcache to 1.3.2

nathanielks:
We weren't able to apply this to `v3-branch`, so manually applying this way.


joehoyle:
@nathanielks I published 2.0.9 in https://github.com/humanmade/altis-cloud/releases/tag/2.0.9, you should be able to `composer update` now to pull in the fix

PR #142, opened 2020-02-24
Update to cavalcade 2.0.0

roborourke:
Cavalcade 2.0.0 features the new pre flight filters, uses a more flexible `get_jobs_by_query()` method, better db indexes and better handling of the installation process.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
 		"humanmade/wp-redis-predis-client": "0.1.0",
 		"humanmade/aws-xray": "1.2.3",
 		"humanmade/s3-uploads": "~2.1",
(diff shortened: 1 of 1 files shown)


joehoyle:
What does it break in BC? (assuming something for the 2.0)

roborourke:
@joehoyle shouldn’t break anything for any standard setup, the 2.0 was more of a reflection the scale of the change to the code base switching to using the filters.

joehoyle:
Ok sounds good!

joehoyle:
Is there a need for a new version of the Runner with this BTW?

roborourke:
No the runner should be fine as is

PR #144, opened 2020-02-26
Reduce polling on CloudWatch in Audit Log results

joehoyle:
Follow up to https://github.com/humanmade/altis-cloud/pull/97.

This limits the request rate to CloudWatch when checking if queries are complete.

### DIFF
--- a/inc/audit_log_to_cloudwatch/class-cloudwatch-driver.php
+++ b/inc/audit_log_to_cloudwatch/class-cloudwatch-driver.php
@@ -142,6 +142,10 @@ class CloudWatch_Driver implements DB_Driver_Interface {
 
 		$results = [ 'status' => 'Running' ];
 		while ( ! in_array( $results['status'], [ 'Failed', 'Cancelled', 'Complete' ], true ) ) {
+			// Limit how fast we poll CloudWatch via calls to getQueryResults.
+			// Queries take at a minimum 1 second, so we `sleep` before even
+			// making the first call.
+			sleep( 1 );
 			$results = cloudwatch_logs_client()->getQueryResults([
 				'queryId' => $query['queryId'],
 			] );


joehoyle:
We need to backport this along with https://github.com/humanmade/altis-cloud/pull/97 to the v2-branch manuall.

inc/audit_log_to_cloudwatch/class-cloudwatch-driver.php
```diff
@@ -142,6 +142,10 @@ public function get_records( $args ) {
 
 		$results = [ 'status' => 'Running' ];
 		while ( ! in_array( $results['status'], [ 'Failed', 'Cancelled', 'Complete' ], true ) ) {
+			// Limit how fast we poll CloudWatch via calls to getQueryResults.
+			// Queries take at a minimum 1 second, so we `sleep` before even
+			// making the first call.
+			sleep( 1 );
```
rmccue: I hate it.

Could we move the initial sleep outside the loop, and decrease the sleeping in the loop? eg 1s + 100ms in the loop

joehoyle: This was intentional, as the API is never sub-secoond fast.

joehoyle: OH wait, I think you re suggesting that, and it doesn't do that right now

joehoyle: But, 100ms is way too little. Queries are in the seconds to 10s of second magnitude.

rmccue: Basically, I'm saying we should have a longer initial wait, and a shorter polling interval, so that we catch it as soon as it finishes without excess delay. eg if it takes 1.01s, wait 1.1s instead of 2s

roborourke:
@joehoyle what's the decision here? I agree with @rmccue's take here but if this is a reasonable stop gap solution until we have a quicker way to query logs (eg. read-only ElasticSearch index) then we should take this improvement over nothing. Even if Ryan hates it.

inc/audit_log_to_cloudwatch/class-cloudwatch-driver.php
joehoyle: This isn't going to reduce the polling much, as the RTT is going to be 100ms+ anyway. Queries are in an order of magnitude slower, so I think 1s is an appropriate resolution.

joehoyle:
I think we have a good resolution (1s) for the length of operation this is. Insights queries are measured in seconds, not milliseconds.

joehoyle:
(need an approval before we can merge)

PR #147, opened 2020-02-26
Add the EC2 Istance ID to AWS Xray traces

joehoyle:
Per the docs at https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-aws

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -486,6 +486,8 @@ function on_request_stats( TransferStats $stats ) {
 function get_ec2_instance_metadata() : array {
 	$has_cache = false;
 	$cache_key = 'altis.ec2_instance_metadata';
+	// Use apcu_* as we only want to store the cache on the current server,
+	// not across all servers (wp_cache_*).
 	$cached_data = apcu_fetch( $cache_key, $has_cache );
 	if ( $has_cache ) {
 		return $cached_data;
@@ -523,6 +525,9 @@ function get_ec2_instance_metadata() : array {
 
 /**
  * Add the EC2 instance data to the Xray root segment.
+ *
+ * This function is called pre-WordPress load, so we don't have access
+ * to all the WordPress functions, hence using Guzzle.
  *
  * @param array $trace
  * @return array
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -3,9 +3,11 @@
 namespace Altis\Cloud;
 
 use const Altis\ROOT_DIR;
+use Exception;
 use function Altis\get_config as get_platform_config;
 use function Altis\get_environment_architecture;
 use function HM\Platform\XRay\on_aws_guzzle_request_stats;
+use GuzzleHttp\Client;
 use GuzzleHttp\TransferStats;
 use HM\Platform\XRay;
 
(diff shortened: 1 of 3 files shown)


inc/namespace.php
```diff
@@ -456,3 +459,92 @@ function on_request_stats( TransferStats $stats ) {
...
+ *    privateIp: string,
+ *    ramdiskId: string,
+ *    region: string,
+ *    version: string,
+ * )
+ */
+function get_ec2_instance_metadata() : array {
+	$has_cache = false;
+	$cache_key = 'altis.ec2_instance_metadata';
+	$cached_data = apcu_fetch( $cache_key, $has_cache );
```
roborourke: For my own understanding why do we need to use `apcu` for this?

Also it looks like the `apcu` package was only just added - should we add a `function_exists()` check here?

inc/namespace.php
```diff
@@ -456,3 +459,92 @@ function on_request_stats( TransferStats $stats ) {
...
+		return [];
+	}
+
+	if ( $request->getStatusCode() !== '200' ) {
+		trigger_error( sprintf( 'Unable to get instance metadata. Returned response code: %s', $request->getStatusCode() ), E_USER_NOTICE );
+		apcu_store( $cache_key, [] );
+		return [];
+	}
+
+	$metadata = json_decode( $request->getBody(), true );
```
roborourke: Looks like you can just do `$request->json();` here

inc/namespace.php
```diff
@@ -456,3 +459,92 @@ function on_request_stats( TransferStats $stats ) {
...
+	if ( ! empty( $trace['parent_id'] ) ) {
+		return $trace;
+	}
+
+	$metadata = get_ec2_instance_metadata();
+	if ( ! $metadata ) {
+		return $trace;
+	}
+
+	$trace['aws']['ec2']['availability_zone'] = $metadata['availabilityZone'];
```
roborourke: It's a bit confusing that we can get both ec2 and ecs instance metadata, I thought they were somewhat exclusive from an infrastructure standpoint. Can you add a few comments to note why there's ec2 data when using ecs / where it comes from? Will there be ec2 metadata when using ecs?

inc/namespace.php
rmccue: This is so it's specifically a per-machine cache I think, whereas the object cache is per-stack.

We should already have had apcu installed I think? We're using it on humanmade-production-v2, so it's been around for a while afaik.

joehoyle: Yeah apcu has been there for some time, we just didn't have it documented. It is indeed to cache per-machine. I'll add an inline comment.

inc/namespace.php
joehoyle: No, ecs runs on ec2, so in the case of running on ecs, we want to send both the underlaying instance data, and the ecs container id.

roborourke: Ah cool, makes sense of course. I thought it was maybe a bit more abstracted than that 👍

joehoyle:
Blocked by https://github.com/humanmade/aws-xray/pull/55

inc/namespace.php
joehoyle: Hmm atleast in the version of Guzzle we have, it looks like that isn't in `MessageInterface`

roborourke: Sad times 😢 As you were...

joehoyle:
Ok unblocked, ready for final review.

inc/namespace.php
```diff
@@ -456,3 +459,97 @@ function on_request_stats( TransferStats $stats ) {
...
+	// not across all servers (wp_cache_*).
+	$cached_data = apcu_fetch( $cache_key, $has_cache );
+	if ( $has_cache ) {
+		return $cached_data;
+	}
+
+	$client = new Client();
+
+	try {
+		$request = $client->request( 'GET', 'http://169.254.169.254/latest/dynamic/instance-identity/document', [
```
roborourke: Any value in documenting where this IP comes from? Assuming it's a standard AWS thing.

rmccue: Yep, it's the standard link-local IP: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

joehoyle: IMO this IP is common enough I think.

PR #148, opened 2020-02-27
WIP: Defer setting constants from production ENV

### LINKED ISSUE #83: Do not autoload wp-config.php
Currently `wp-config.php` is autoloaded via composer's autoload files feature. This is included prior to the `altis.loaded_autoloader` hook which is when `.config/load.php` is included.

The end result is that default constants set in `wp-config-production.php` on our stacks can never be overridden if needed and other workarounds using filters are required.

We should load `wp-config.php` intentionally on the `altis.loaded_autoloader` hook with a higher priority than `.config/load.php`.

cc @stuartshields @nathanielks @Japh

roborourke:
Fixes #83

Currently we autoload a file that sets constants from everything in `$_ENV` - this happens before `.config/load.php` is included so its impossible to override things like the `S3_BUCKET_URL` etc which needs to be done for some set ups.


roborourke:
Think this one will need to be tested out on the demo site before we merge to be really sure it's ok.

joehoyle:
Yeah I’m pretty sure this is going to break a bunch of stuff. Currently one expects to be able to get things like the current environment type etc in the load.php, and this will change that I believe. This has come up in a couple of other places, and I think what we need is another “load-early.php” or similar, because there is no “right answer”. Sometimes you want to rub stuff before defaults have been set up, and sometimes after.

As it happens you can already do a “pre-load” file by setting up an auto load.files in the root composer.json, so this could be just a case of documenting that?

joehoyle:
Yeah I’m pretty sure this is going to break a bunch of stuff. Currently one expects to be able to get things like the current environment type etc in the load.php, and this will change that I believe. This has come up in a couple of other places, and I think what we need is another “load-early.php” or similar, because there is no “right answer”. Sometimes you want to rub stuff before defaults have been set up, and sometimes after.

As it happens you can already do a “pre-load” file by setting up an auto load.files in the root composer.json, so this could be just a case of documenting that?

roborourke:
Ah yeah forgot about env type etc... hmm

> As it happens you can already do a “pre-load” file by setting up an auto load.files in the root composer.json, so this could be just a case of documenting that?

Will that reliably be included first?

Alternatively could we be more specific in `wp-config-production.php` rather than looping over everything?

Looking at the new `wp-config-production.php` there's nothing dynamic in there any more so we could move that whole thing into this repo anyway.

So if `getenv( 'HM_ENV_TYPE' )` is set we define all the constants we need early and don't want people to be able to override in `load.php`, then define anything left undefined in `$_ENV` rest on the hook proposed here?

roborourke:
Noodling around with the idea bit more - still WIP but we could go this direction. It might even be better not to loop over `$_ENV` at all but to write out the env -> constant mapping in full to really communicate what gets set from where and when.

nathanielks:
I'm personally more in favor with adding a file that gets loaded early as @joehoyle suggests than changing this, but if you wanted to keep pursuing this idea... 👇 

> Alternatively could we be more specific in wp-config-production.php rather than looping over everything?

I'm not usually a fan of "magic," but this is one case where I am a fan. It adds a lot of flexibility and reduces friction when wanting to get ideas out the door by simply  defining values set from the environment. 

For what's currently implemented, there are a number of values that wouldn't be caught by what you have, though. Here are some more variables that get passed:

```
[
  "ALTIS_ANALYTICS_COGNITO_ID",
  "ALTIS_ANALYTICS_COGNITO_REGION",
  "ALTIS_ANALYTICS_PINPOINT_ID",
  "ALTIS_ANALYTICS_PINPOINT_REGION",
  "AWS_SES_WP_MAIL_REGION",
  "AWS_SES_WP_MAIL_USE_INSTANCE_PROFILE",
  "AWS_XRAY_DAEMON_IP_ADDRESS",
  "DB_HOST",
  "DB_NAME",
  "DB_READ_REPLICA_HOST",
  "DB_USER",
  "ELASTICSEARCH_HOST",
  "ELASTICSEARCH_PORT",
  "HM_ENV",
  "HM_ENV_ARCHITECTURE",
  "HM_ENV_REGION",
  "HM_ENV_TYPE",
  "MEMORY_LIMIT",
  "REDIS_HOST",
  "REDIS_PORT",
  "S3_UPLOADS_BUCKET",
  "S3_UPLOADS_BUCKET_URL",
  "S3_UPLOADS_REGION",
  "S3_UPLOADS_USE_INSTANCE_PROFILE",
  "STACK_NAME"
]
```

From what I can tell, you'd need to add if you wanted to keep going with this:
```
populate_constants_from_env( 'DB_' );
populate_constants_from_env( 'ELASTICSEARCH_' );
populate_constants_from_env( 'REDIS_' );
populate_constants_from_env( 'S3_' );
```

IIRC `HMN_ENV` is the same as `STACK_NAME`, so that's not 100% necessary to duplicate.

roborourke:
@nathanielks cheers for the list!

> I'm not usually a fan of "magic," but this is one case where I am a fan. It adds a lot of flexibility and reduces friction when wanting to get ideas out the door by simply defining values set from the environment.

Yeah I thought that was the case, it does make sense.

> For what's currently implemented, there are a number of values that wouldn't be caught by what you have, though.

So on these lines I'm calling `populate_constants_from_env()` without the prefix argument - that will catch and define any env vars that haven't yet been defined. I'm doing this because the problem I'm hoping to solve is to allow custom config from `.config/load.php` to be able to modify a subset of the constants.

https://github.com/humanmade/altis-cloud/blob/1f44ad72c6f2403394c1bb13134dc8d132a0c939/load.php#L67-L69

The `load.php` file in this case does get called really early (immediately after `vendor/autoload.php` in `wp-config.php`) before the modules are actually bootstrapped so this is effectively that best of both worlds solution.

It actually looks `HM_ENV_` and `AWS_` are really the only super important prefixes to define early.

Basically I'm not into the idea of relying on composer's autoload files to make sure yet-another-config-file is loaded before `wp-config-production.php` and having to document another special case for adding config.

nathanielks:
> that will catch and define any env vars that haven't yet been defined. 

Aha! I totally missed that, thanks!

roborourke:
Going to close this out - it's complex and this solution still doesn't being clarity as to what constants are defined by environment variables and which aren't.

A better solution will be to push the use of the environment variables lower to where we currently use constants such as in the core `get_environment_type()` etc functions.

PR #150, opened 2020-03-03
Update AWS X-Ray to 1.2.5

rmccue:
Fixes metadata (https://github.com/humanmade/aws-xray/pull/56) and segment details in Query Monitor (https://github.com/humanmade/aws-xray/pull/51).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 	"require": {
 		"php": ">=7.1",
 		"humanmade/wp-redis-predis-client": "0.1.0",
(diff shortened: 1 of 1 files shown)


joehoyle:
Think this should be 1.2.6 now?

rmccue:
If you're on board with it, we can switch to the `~` requirement here instead

joehoyle:
Yeah thikn that would be ok

roborourke:
Added the `~` constraint.

PR #158, opened 2020-03-14
Changing to __healthcheck, changing json response

### LINKED ISSUE #135: Healthcheck not available in JSON
We are intending to support fetching the `/healthcheck/` is JSON format too, by either specifying `Accept: appliction/json` or adding an `_accept=json` url param. Because CloudFront doesn't forward `Accept`, when wanting to fetch the Healthcheck externally, you'd expect to pass `?_accept=json` -- however that doesn't work because the condition to check the healthcheck is then not matched in https://github.com/humanmade/altis-cloud/blob/master/inc/healthcheck/namespace.php#L13 (`'/healthcheck/' !== $_SERVER['REQUEST_URI']`) ( `$_SERVER['REQUEST_URI']` includes query params).

We should be matching `/healthcheck/` by some means where query params are stripped. IMO while we are at it -- allow `/healthcheck` (no trailing slash) to also match.

---------------

Acceptance criteria:

- [ ] Healthcheck can be accessed via `/__healthcheck` or `/__healthcheck/` with any query string
- [ ] If the header `Accept` has the value `application/json` return a JSON response
- [ ] If the query string contains `_accept=json` return a JSON response
- [ ] The response should look like the following:
  ```json
  {
	"status": "failed | ok",
	"checks" : {
		"mysql": {
			"status": "ok | failed",
			// only add the error if the check returned a WP_Error
			"error": {
				"code": "error code",
				"message": "error message"
			}
		},
		"...": "..."
	}
  }
  ```

ivdimova:
Fixes #135

### DIFF
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -10,8 +10,8 @@ function bootstrap() {
 		require_once __DIR__ . '/class-cli-command.php';
 		WP_CLI::add_command( 'healthcheck', __NAMESPACE__ . '\\CLI_Command' );
 	}
-
-	if ( ! isset( $_SERVER['REQUEST_URI'] ) || strpos( $_SERVER['REQUEST_URI'], '__healthcheck' ) == false ) {
+	
+	if ( ! isset( $_SERVER['REQUEST_URI'] ) || strpos( $_SERVER['REQUEST_URI'], '/__healthcheck' ) !== 0 ) {
 		return;
 	}
 	output_page( run_checks() );
@@ -41,7 +41,7 @@ function output_page( array $checks ) {
 	}
 
 	$json_response = [
-		'status' => $passed ? 'OK' : 'Failure!',
+		'status' => $passed ? 'ok' : 'failed',
 		'checks' => $checks,
 	];
 
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -10,7 +10,7 @@ function bootstrap() {
 		require_once __DIR__ . '/class-cli-command.php';
 		WP_CLI::add_command( 'healthcheck', __NAMESPACE__ . '\\CLI_Command' );
 	}
-	
+
 	if ( ! isset( $_SERVER['REQUEST_URI'] ) || strpos( $_SERVER['REQUEST_URI'], '/__healthcheck' ) !== 0 ) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


inc/healthcheck/namespace.php
```diff
@@ -10,7 +10,8 @@ function bootstrap() {
 		require_once __DIR__ . '/class-cli-command.php';
 		WP_CLI::add_command( 'healthcheck', __NAMESPACE__ . '\\CLI_Command' );
 	}
-	if ( ! isset( $_SERVER['REQUEST_URI'] ) || '/healthcheck/' !== $_SERVER['REQUEST_URI'] ) {
+
+	if ( ! isset( $_SERVER['REQUEST_URI'] ) || strpos( $_SERVER['REQUEST_URI'], '__healthcheck' ) == false ) {
```
rmccue: This should check that it's at the start of the URL, not just in it, otherwise we might hijack too many URLs. Also, should use a strict check (`===`).
```suggestion
	if ( ! isset( $_SERVER['REQUEST_URI'] ) || strpos( $_SERVER['REQUEST_URI'], '/__healthcheck' ) === 0 ) {
```

inc/healthcheck/namespace.php
```diff
@@ -39,9 +40,14 @@ function output_page( array $checks ) {
 		$format = 'json';
 	}
 
+	$json_response = [
+		'status' => $passed ? 'OK' : 'Failure!',
```
rmccue: I believe the idea was for this to be machine-readable, so should be a status code rather than human text:
```suggestion
		'status' => $passed ? 'ok' : 'failed',
```

ivdimova: I missed this - fixed.

inc/healthcheck/namespace.php
ivdimova: Hmm, ```Healthcheck can be accessed via /__healthcheck or /__healthcheck/ with any query string``` - I thought it should be accessible everywhere?

rmccue: Only at the root (e.g. `humanmade.com/__healthcheck`), but any query string should be allowed (e.g. `humanmade.com/__healthcheck?whatever`)

ivdimova: I am not sure why I understood it otherwise. Done.

ivdimova:
@roborourke @rmccue I assume this means I can merge and close the issue?

roborourke:
@ivdimova  correct! Thanks for this

PR #163, opened 2020-03-23
Issue 87/subsite cron healtchech

ivdimova:


joehoyle:
@ivdimova thanks for the PR! Could you make sure you use titles to PRs formatted along the lines of other PRs at https://github.com/humanmade/altis-cloud/pulls, as the titles are used in the changelog. They should use succict sentence formatted English.

Do you know what is limiting the jobs being created on sub-sites? I noticed in https://github.com/humanmade/altis-cloud/blob/master/inc/healthcheck/cavalcade/namespace.php#L16 that there's no check if "only register the job if it's on the main site?

inc/healthcheck/cavalcade/namespace.php
```diff
@@ -51,10 +51,13 @@ function set_last_run() {
  * @return boolean|WP_Error True if healthy, error otherwise.
  */
 function check_health() {
+	switch_to_blog( 1 );
```
joehoyle: Hmm I'm not sure if `1` is always the case, I think this should use whatever is checked to actually enqueue the job in the first place.

ivdimova:
> @ivdimova thanks for the PR! Could you make sure you use titles to PRs formatted along the lines of other PRs at https://github.com/humanmade/altis-cloud/pulls, as the titles are used in the changelog. They should use succict sentence formatted English.
> 
> Do you know what is limiting the jobs being created on sub-sites? I noticed in https://github.com/humanmade/altis-cloud/blob/master/inc/healthcheck/cavalcade/namespace.php#L16 that there's no check if "only register the job if it's on the main site?

@joehoyle - I just started looking at this today, so I am not sure why it's limited but seems this is the line that does it? https://github.com/humanmade/altis-cloud/blob/master/inc/healthcheck/cavalcade/namespace.php#L22

inc/healthcheck/cavalcade/namespace.php
ivdimova: This seem to work?

joehoyle: What if the main site ID isn't `1` though?

ivdimova: I did change it last night, please see the latest commit @joehoyle ?

inc/healthcheck/cavalcade/namespace.php
```diff
@@ -51,10 +51,13 @@ function set_last_run() {
  * @return boolean|WP_Error True if healthy, error otherwise.
  */
 function check_health() {
+	switch_to_blog( get_network()->id );
```
joehoyle: This get's the network's ID, not the main blog on the network. In WordPress you can have many networks, with many blogs in each network. Confusingly "site" can either mean "blog" or "network" depending on when the function was written.

We only enqueue this job when `is_main_site()` is `true`, `is_main_site` uses `get_main_site_id()` (site in this context means "blog"), so this should use `switch_to_blog( get_main_site_id() );`

We do have another bug here though, because if the site has many networks (which isn't default with WordPress, but is possible to set up), we'll be registering one job per network (on the main site of each network). I don't thikn we need to fix that in this PR, but we should make a ticket, cc @rmccue who wrote the intial cavalcade job checker, does that make sense?

ivdimova: Good point - I considered `get_main_site_id()` but wasn't sure if the network ID is not better - obviously the confusion comes from the way these are called. Just changed it.

rmccue: @joehoyle Yeah, I think we probably want to switch to the main blog of the main network, which would be `switch_to_blog( get_network( get_main_network_id() )->site_id )`

ivdimova: @rmccue @joehoyle so finally which one is correct?

joehoyle: What you changed it to is good! We’ll want to fix that other issue at some point but doesn’t need to happen in this pr

PR #164, opened 2020-04-03
Only call fastcgi_finish_request on shutdown

joehoyle:
It turns out when the CloudWatch Error Handler was active, we'd basically call `fastcgi_finish_request` as soon as > 100 errors were caught. This meant requests could randomly return 200 responses with no body.

This is an important fix to backport as far back as we can!

### DIFF
--- a/inc/cloudwatch_error_handler/namespace.php
+++ b/inc/cloudwatch_error_handler/namespace.php
@@ -24,7 +24,7 @@ function bootstrap() {
 		return false;
 	} );
 
-	register_shutdown_function( __NAMESPACE__ . '\\send_buffered_errors' );
+	register_shutdown_function( __NAMESPACE__ . '\\send_buffered_errors_on_shutdown' );
 }
 
 function error_handler( int $errno, string $errstr, string $errfile = null, int $errline = null ) : bool {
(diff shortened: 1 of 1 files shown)


inc/cloudwatch_error_handler/namespace.php
```diff
@@ -48,6 +48,24 @@ function error_handler( int $errno, string $errstr, string $errfile = null, int
...
+	if ( function_exists( 'fastcgi_finish_request' ) ) {
+		fastcgi_finish_request();
+	}
+	send_buffered_errors();
+}
+
+/**
+ * Send all buffered errors to CloudWAtch.
+ */
 function send_buffered_errors() {
```
igmoweb, line 69: Is this function not called anymore?

joehoyle: Yup it's called right above, and still called from `error_handler` too.

igmoweb: Ah, yeah, I missed the last line

inc/cloudwatch_error_handler/namespace.php
```diff
@@ -48,6 +48,24 @@ function error_handler( int $errno, string $errstr, string $errfile = null, int
 	return false;
 }
 
+/**
+ * When script execution ends, send any buffered errors to CloudWatch.
+ */
+function send_buffered_errors_on_shutdown() {
+	$errors = $GLOBALS['altis_cloudwatch_error_handler_errors'];
```
rmccue, line 55: Is this always set?

joehoyle: Mm I'll add a guard, just in case. We don't want a recursive error!

inc/cloudwatch_error_handler/namespace.php
```diff
@@ -48,6 +48,23 @@ function error_handler( int $errno, string $errstr, string $errfile = null, int
...
+ * When script execution ends, send any buffered errors to CloudWatch.
+ */
+function send_buffered_errors_on_shutdown() {
+	if ( empty( $GLOBALS['altis_cloudwatch_error_handler_errors'] ) ) {
+		return;
+	}
+
+	if ( function_exists( 'fastcgi_finish_request' ) ) {
+		fastcgi_finish_request();
+	}
```
rmccue, line 61: By the way, is this never not defined for any of our environments?

Issue #159, opened 2020-03-16
Add instance healthcheck to be used for ALB healthcheck


joehoyle:
Currently our ALB (or specifically Target Group) healthcheck sends requests to `wp-admin/install.php` to check for 200 responses. This is a healthcheck to determine if a specific instance is healthy, not the whole application (which is what `/__healthcheck` is for).

The _instance healthcheck_ needs to determine if PHP is functioning correctly on the specific instance, and currently, not much else. Rather than rely on `wp-admin/install.php` for this, which is not built this for this purpose, we should introduce a new path of `/__instance_healthcheck` that works similat to the current `/__healthcheck` but only checks things that could be a problem with the current server. Currently that would just be "return true", because if PHP has managed to execute our `/__instance_healthcheck`, then that's about all the checking we currently need.

In the future we may want to add other checks too, so we should use the same code patterns as `/__healthcheck` and I'd recommend we just co-locate the code for both in the same place.

I think it would be good to also exclude requests to this URL from AWS Xray as they are not public web requests, and often get in the way when trying to debug things.

cc @humanmade/servers if this sounds appropriate, as once this change is in Altis, we'll want to update terraform-app-stack too.

-------------

Acceptance criteria:

- [ ] The URL `/__instance_healthcheck` or `/__instance_healthcheck/` should return an empty response
- [ ] The HTTP response code should be 200

nathanielks:
Yup, mostly sounds good to me 👍 With not including X-Ray, so I'd like to ensure we have verbose error logging in that endpoint (should it ever include more than `return true;`.

roborourke:
@joehoyle I updated this with some acceptance criteria - can you confirm that's correct and edit if needed?

joehoyle:
Looks good!

ivdimova:
I see the `healthcheck` works as a `wp cli` command as well, should there be another command for `instance healthcheck`?

joehoyle:
@ivdimova good catch, i think we might as well for constistency yeah! 👍

roborourke:
Done in #171

PR #171, opened 2020-04-24
Add Instance Healthcheck

abhishek-kaushik:
Add feature for #159

### DIFF
--- a/inc/healthcheck/class-cli-command.php
+++ b/inc/healthcheck/class-cli-command.php
@@ -45,6 +45,8 @@ class CLI_Command extends WP_CLI_Command {
 	 * @param array $args
 	 * @param array $args_assoc
 	 *
+	 * @since 4.0
+	 *
 	 * @return void
 	 */
 	function instance( array $args, array $args_assoc ) {
--- a/inc/healthcheck/class-cli-command.php
+++ b/inc/healthcheck/class-cli-command.php
@@ -11,10 +11,18 @@ class CLI_Command extends WP_CLI_Command {
 	/**
 	 * Run health checks.
 	 *
-	 * @param array $args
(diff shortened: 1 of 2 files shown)


inc/healthcheck/class-cli-command.php
```diff
@@ -7,6 +7,15 @@
...
 
 class CLI_Command extends WP_CLI_Command {
+
+	/**
+	 * Run health checks.
+	 *
+	 * @param array $args
+	 * @param array $args_assoc
+	 *
+	 * @return void
```
joehoyle, line 17: No need for the void return hinting. I believe these should also use `@synopsis` WP CLI documentation aswell than params

inc/healthcheck/class-cli-command.php
```diff
@@ -29,4 +38,39 @@ function run( array $args, array $args_assoc ) {
...
 		WP_CLI::halt( $passed ? 0 : 1 );
 	}
+
+	/**
+	 * Run instance health checks.
+	 *
+	 * @param array $args
+	 * @param array $args_assoc
+	 *
+	 * @since 4.0
```
joehoyle, line 48: We don't need the version, we don't use that else.

inc/healthcheck/namespace.php
```diff
@@ -83,12 +93,29 @@ function run_checks() : array {
...
+ */
+function run_instance_checks() : array {
+	$checks = [
+		'php' => true,
+	];
+
+	/**
+	 * Filters Instance Healthchecks response.
+	 *
+	 * @since 4.0
```
joehoyle, line 114: No need for since, and needs docs for the $checks param

joehoyle:
Thanks @abhishek-kaushik !

PR #172, opened 2020-05-01
Fix page cache to allow dynamic unique param

faishal:
This will allow developer to add `unique` based on query string or any dynamic conditions.

We can also add filter for the same.

### DIFF
--- a/inc/page_cache/namespace.php
+++ b/inc/page_cache/namespace.php
@@ -7,7 +7,10 @@ use function Altis\Cloud\get_config;
 
 function bootstrap() {
 	global $batcache;
-	$batcache = $batcache ?? [ 'ignored_query_string_params' => [], 'unique' => [] ];
+	$batcache = $batcache ?? [
+		'ignored_query_string_params' => [],
+		'unique' => []
+	];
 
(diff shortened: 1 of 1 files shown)


joehoyle:
Hmm it's not clear to me how this works. I'm not sure if this is conflating `get_ignore_query_params` with `unique`, as those two things are "opposites". `get_ignore_query_params` allows you to say `page doesn't need to be unique for these query params`. Is this intending to add a `unique` api in the `composer.json`, or is it just supporting people setting `unqiue` on the batcache global?

roborourke:
@joehoyle we have a way to set both unique headers and ignored query vars via composer - this is addressing the ability to set them dynamically in code before Altis kicks in. You can't do conditional checks in `composer.json` essentially so this is to address that.

It isn't merging the unique and ignored query strings - just defaulting them both to empty arrays as keys on the `$batcache` global.

roborourke:
Actually @faishal can you document how this would be used as well?

joehoyle:
Ahh ok cool, so this is a regression for people who were usign `global $batcache` I support?

roborourke:
Yep, exactly that

PR #173, opened 2020-05-05
Add wp-db bail override

joehoyle:
Currently `wpdb::bail` will not `exit` if `wpdb->show_errors` is set to `false`. This can be the case in several situations, such as Query Monitor being enabled. Also, the default `wpdb::bail` implementation use `wp_die` which reads form the database, causing a circular call loop in some cases.

We don't need to ever no-exit on a database bail call.

### DIFF
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -137,7 +137,7 @@ class DB extends LudicrousDB {
 	 * This function can not use any WordPress functions that read from the database, as that
 	 * will risk a recursion call.
 	 *
-	 * @param string $message    The Error message
+	 * @param string $message The Error message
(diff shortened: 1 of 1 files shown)


inc/class-db.php
```diff
@@ -129,4 +129,28 @@ public function add_database( array $db = [] ) {
...
 	}
+
+	/**
+	 * Override the wpdb bail handler.
+	 *
+	 * We always want to bail and exit the script execution, not rely on $this->show_errors.
+	 * This function can not use any WordPress functions that read from the database, as that
+	 * will risk a recursion call.
+	 *
+	 * @param string $message    The Error message
```
rmccue, line 140:
```suggestion
	 * @param string $message The Error message
```

PR #157, opened 2020-03-10
Cloudfront media purge

faishal:
- Creates a CloudFront invalidation request for deleted media on delete attachment.

PR #177, opened 2020-05-20
Backport Cloudfront media purge v3

joehoyle:
See https://github.com/humanmade/altis-cloud/pull/157

### DIFF
--- a/docs/cdn-purge.md
+++ b/docs/cdn-purge.md
@@ -3,8 +3,8 @@
 The majority of the requests, data served by Altis will be cached on the CDN to improve performance and delivery time to the user. In some situations you may need to invalidate specific URLs on the CDN. You can add your own CDN purge rule directly in PHP using `Altis\Cloud\purge_cdn_paths()` helper function.
 
 ```php
-use Altis\Cloud\purge_cdn_paths;
-purge_cdn_paths( [
+use Altis\Cloud;
+Cloud\purge_cdn_paths( [
    '/sample-page/*',
 ] );
 ```
--- a/docs/cdn-purge.md
+++ b/docs/cdn-purge.md
@@ -1,0 +1,32 @@
+# Purging Cache on the CDN
+
+The majority of the requests, data served by Altis will be cached on the CDN to improve performance and delivery time to the user. In some situations you may need to invalidate specific URLs on the CDN. You can add your own CDN purge rule directly in PHP using `Altis\Cloud\purge_cdn_paths()` helper function.
+
+```php
+use Altis\Cloud\purge_cdn_paths;
+purge_cdn_paths( [
+   '/sample-page/*',
+] );
+```
+
+Note that purging paths should be kept to only essential items as purging more than 1000 paths per month will incur overage charges.
+
+## Automatic media purge rule
+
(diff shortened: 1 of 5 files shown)


joehoyle:
@stuartshields this is the PR to backport to v3. You don't probably want to use the Alti config, but this will give you access to `Altis\Cloud\Cloudfront_Media_Purge\purge_media_file_cache` and `Altis\Cloud\Cloudfront_Media_Purge\purge_cdn_paths`.

docs/cdn-purge.md
```diff
@@ -0,0 +1,32 @@
+# Purging Cache on the CDN
+
+The majority of the requests, data served by Altis will be cached on the CDN to improve performance and delivery time to the user. In some situations you may need to invalidate specific URLs on the CDN. You can add your own CDN purge rule directly in PHP using `Altis\Cloud\purge_cdn_paths()` helper function.
```
roborourke, line 3:
```suggestion
For the majority of the requests, data served by Altis will be cached on the CDN to improve performance and delivery time to the user. In some situations you may need to invalidate specific URLs on the CDN. You can add your own CDN purge rule directly in PHP using the `Altis\Cloud\purge_cdn_paths()` helper function.
```

docs/cdn-purge.md
```diff
@@ -0,0 +1,32 @@
...
+
+```php
+use Altis\Cloud\purge_cdn_paths;
+purge_cdn_paths( [
+   '/sample-page/*',
+] );
```
roborourke, line 9: Just realised this is invalid PHP, we should fix it upstream too:
```suggestion
use Altis\Cloud;
Cloud\purge_cdn_paths( [
   '/sample-page/*',
] );
```

joehoyle:
Yup agreed, thanks for the review

stuartshields:
@joehoyle I believe this backport is missing:
```
use function Altis\get_aws_sdk;
```

We'd need to also update the other backport as well.

inc/namespace.php
```diff
@@ -424,3 +436,70 @@ function remove_xray_metadata( array $metadata ) : array {
...
+	if ( defined( 'CLOUDFRONT_DISTRIBUTION_ID' ) ) {
+		$distribution_id = CLOUDFRONT_DISTRIBUTION_ID;
+	}
+
+	/**
+	 * Filters the CloudFront Distribution ID used when purging CDN paths.
+	 *
+	 * @param string $distribution_id The ID to set.
+	 */
+	$distribution_id = apply_filters( 'altis.cloud.cdn_distribution_id', $distribution_id );
```
stuartshields, line 479: @joehoyle I'd recommend adding something to the docs about this filter, if it hasn't already. If you add `get_aws_sdk` from https://github.com/humanmade/altis-core/blob/master/docs/README.md#aws-sdk this will fatal locally.

joehoyle: Hmm yeah, you need to make sure any code calling this is wrapped in checks that it's running in Cloud

joehoyle:
Thanks @stuartshields, did you deploy this to verify that? If so, I know I can merge without risk :)

stuartshields:
@joehoyle So I was testing this on dev and have not run into any issues.

joehoyle:
@stuartshields ok this has been backported to https://github.com/humanmade/altis-cloud/releases/tag/3.0.5

PR #180, opened 2020-06-05
Fix linter errors

roborourke:
Fixes docblocks and adds file docblocks. This should prevent problems as our standards become stricter in future. It also makes the code style more consistent regarding alignment of array keys and parameter comments.


joehoyle:
Nice on @roborourke

PR #185, opened 2020-06-11
Enable configuring email region

### LINKED ISSUE #149: Add configuration for email sending region
In most cases, the email sending region automatic selection works fine. However, when reusing a domain from an environment in a different region, it can be useful to manually override the email region for SES.

For example, if I have an environment in us-east-1 already sending email from `example.com`, and I'm setting up another environment in eu-west-1 using the same email domain, I should be able to send email using us-east-1.

This avoids requiring duplicate domain verification for each region (which can take a while to pass through various security and compliance checks).

Additionally, it allows us to potentially add more regions to SES in the future as AWS adds them, while allowing setting the region manually to maintain backwards compatibility.

To achieve this, we need to allow overriding the `AWS_SES_WP_MAIL_REGION` constant before `wp-config-production.php` is loaded. This may require fixing #83 first.

Acceptance criteria:
* [x] When I set the `email-region` option to a region string (such as `us-east-1`), email should be sent from the corresponding AWS region
* [x] When I don't set the `email-region` option, email should be sent from the region specified in `wp-config-production.php`
* [x] Given I'm already sending email from an environment, when I upgrade, email should be sent from the same region

roborourke:
Fixes #149

### DIFF
--- a/docs/email-delivery.md
+++ b/docs/email-delivery.md
@@ -42,7 +42,7 @@ Any email address can be used as long as they are all from a verified sending do
 
 In the event of migrating from one region to another or setting up AWS SES in a region other than the one your application is hosted in you may need to override the default region.
 
-To configure the sending region set the `email-region` configuration option to a valid AWS region code as shown in the following example.
+To configure the sending region set the `email-region` configuration option to a valid AWS region code as shown in the following example. (By default this is set to `false`, which uses the default region.)
 
 ```json
 {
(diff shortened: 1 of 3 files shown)


docs/email-delivery.md
```diff
@@ -36,6 +38,22 @@ add_filter( 'wp_mail_from_name', function ( $name ) {
 
 Any email address can be used as long as they are all from a verified sending domain. Multiple sending domains can also be used, just get each one verified.
 
+## Email Sending Region
+
+In the event of migrating from one region to another or setting up AWS SES in a region other than the one your application is hosted in you may need to override the default region.
+
+To configure the sending region set the `email-region` configuration option to a valid AWS region code as shown in the following example.
```
rmccue, line 45:
```suggestion
To configure the sending region set the `email-region` configuration option to a valid AWS region code as shown in the following example. (By default this is set to `false`, which uses the default region.)
```

PR #188, opened 2020-06-18
Move ES request signing to Cloud Module

### LINKED ISSUE #187: Move ElasticSearch request signing from Search Module to Cloud Module
Right now if the search module is not enabled Altis will fail to sign requests to ES for analytics requests.

This should be handled by the cloud module as ES is not solely used by the search module any more.

roborourke:
ElasticSearch is no longer solely used for search, rather it is a core service of the infrastructure and therefore the cloud module.

This PR replicates the request signing, error logging and healthcheck from the search module in a back compatible way.

Currently if the search module is not enabled then ES requests for analytics data for example will fail. This is definitely a bug that needs to be addressed.

Fixes #187

### DIFF
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -7,6 +7,7 @@
 
 namespace Altis\Cloud\Healthcheck;
 
+use Altis\Cloud;
 use WP_CLI;
 use WP_Error;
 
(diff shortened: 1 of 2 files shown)


joehoyle:
@roborourke this seems a bit risky to backport -- or rather, will this be fine working with the old version of the search modue?

roborourke:
@joehoyle yeah it should be ok but I'll double check. That specific code in the search module hasn't really changed since v1.

Issue #186, opened 2020-06-15
Queue purge media cache when rate limited.


stuartshields:
@joehoyle So I ran into an issue that will cause issues for other projects I believe with using:
https://github.com/humanmade/altis-cloud/blob/master/inc/cloudfront_media_purge/namespace.php#L29

```
TooManyInvalidationsInProgress (client): Processing your request will cause you to exceed the maximum number of in-progress wildcard invalidations. 
```

@Japh suggested that we need to queue this so we aren't rate limited by AWS.

Thoughts?

stuartshields:
We also get the following as well not sure if related:
```
InvalidArgument (client): Your request contains a caller reference that was used for a previous invalidation batch for the same distribution.
```

joehoyle:
@stuartshields it looks like you might be sending too many invalidation requests at once. When you want to purge an attachment, are you doing one purge request per file, or passing them all at once?

I'll look into what the rate limit is too.

joehoyle:
Per the AWS docs, this is the limits:

> If you’re invalidating files individually, you can have invalidation requests for up to 3,000 files per distribution in progress at one time. This can be one invalidation request for up to 3,000 files, up to 3,000 requests for one file each, or any other combination that doesn’t exceed 3,000 files. For example, you can submit 30 invalidation requests that invalidate 100 files each. As long as all 30 invalidation requests are still in progress, you can’t submit any more invalidation requests. If you exceed the maximum, CloudFront returns an error message.

> If you’re using the * wildcard, you can have requests for up to 15 invalidation paths in progress at one time. You can also have invalidation requests for up to 3,000 individual files per distribution in progress at the same time; the maximum on wildcard invalidation requests allowed is independent of the maximum on invalidating files individually. 

@stuartshields could you detail how many concurrent invalidations you are trying to do? Is this "normal operation", or are you running a migration or something that's causing a lot of requests?

stuartshields:
@joehoyle So I found an issue that when it was saving it was hitting the AWS function 4~ times. I think this will solve the issue somewhat. 

Will we hit the rate limit if the post has say 20 attachments?

joehoyle:
@stuartshields Hmm if you are using `purge_media_file_cache()` yes, that will have a concurrency limit of 15 I think. You could use `Altis\Cloud\purge_cdn_paths()` directly (if you are not now) to invalidate up to 3k absolute paths too, but I think the wildcard approach is probably more what you want.

[...]

PR #196, opened 2020-06-19
Document limitation of purge API

rmccue:
See #186.


rmccue:
We agreed in-principle on the changes in #197 instead, so closing in favour of that (which contains these commits).

PR #197, opened 2020-06-19
Enforce purge limitation

rmccue:
Replaces #196. See #186.

This goes further and enforces a maximum of 10 items per HTTP request. This is designed to be lower than the wildcard limit (15) in CloudFront, and is mainly designed to detect use of the function in loops, which is dangerous.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -20,6 +20,13 @@ use HM\Platform\XRay;
 use Psr\Http\Message\RequestInterface;
 
 /**
+ * CloudFront invalidation limit.
+ *
+ * @see purge_cdn_paths()
+ */
+const INVALIDATION_LIMIT = 10;
+
+/**
  * Set up the Cloud Module.
  */
 function bootstrap() {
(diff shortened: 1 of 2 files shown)


rmccue:
@joehoyle that was intentional; for full paths, it seems like they should still be contacting us?

joehoyle:
@rmccue no I don’t think so, it would be fine to do 2k full paths as part of a wp cli command, say

roborourke:
Can we get this finished up today @rmccue ? I can take a look if you're stacked

inc/namespace.php
```diff
@@ -748,6 +772,28 @@ function get_cloudfront_client() : CloudFrontClient {
  * @return bool Returns true if invalidation successfully created, false on failure.
  */
 function purge_cdn_paths( array $paths_patterns ) : bool {
+	static $invalidated_static = 0;
+	static $invalidated_wild = 0;
+	foreach ( $paths_patterns as $pattern ) {
+		if ( strpos( $pattern, '*' ) !== null ) {
```
joehoyle, line 778: `strpos` returns false not null

rmccue: blech, I'm an idiot

rmccue:
Fixed, and re-tested locally this time.

PR #200, opened 2020-06-25
Disable persistent connections to prevent idle resources from being consumed

nathanielks:
We've observed some environments have a high level of connections that are sitting idle, which wastes resources and triggers false alarms. I'd like to experiment with disabling persistent connections to see what difference that makes.

### DIFF
--- a/load.php
+++ b/load.php
@@ -9,6 +9,8 @@ namespace Altis\Cloud; // phpcs:ignore
 
 use Altis;
 
+defined( 'ALTIS_PERSISTENT_DB_CONNS' ) or define( 'ALTIS_PERSISTENT_DB_CONNS', true );
+
 add_action( 'altis.modules.init', function () {
 	$is_cloud = in_array( Altis\get_environment_architecture(), [ 'ec2', 'ecs' ], true );
 	$default_settings = [
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -30,7 +30,7 @@ class DB extends LudicrousDB {
 	 *
 	 * @var bool
 	 */
-	public $persistent = true;
+	public $persistent = false;
 
 	/**
 	 * Track total time waiting for database responses;
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -30,7 +30,7 @@ class DB extends LudicrousDB {
 	 *
 	 * @var bool
 	 */
-	public $persistent = false;
+	public $persistent = ALTIS_PERSISTENT_DB_CONNS;
 
 	/**
 	 * Track total time waiting for database responses;


load.php
```diff
@@ -9,6 +9,8 @@
 
 use Altis;
 
+defined( 'ALTIS_PERSISTENT_DB_CONNS' ) or define( 'ALTIS_PERSISTENT_DB_CONNS', true );
```
nathanielks, line 12: I reckon there's another place I can/should define this, so I'm open.

joehoyle:
I'm not against merging this as it's not changing the default, though I don't really understand the hypothesis, or what is to be learned from doing this.

I see two key questions to this:

1. Does the open connection count on the MySQL servers(s) match the total number of running PHP-FPM workers? I'd expect that it should -- it might also be possible that the open connection count is more like theoretical limit of PHP-FPM workers, because though workers are shutdown when idle for an amount of time, I don't know if PHP-FPM will also reduce the mysql connection pool of persistent connections.

Assuming 1 holds, which I'd fully expect it to, then I'd expect this behavior (idle mysql connections) (which btw I don't think does use any resources while idle, do you have anything to point to the contrary?). In which case, what are you expecting to show by disabling the persistent connections? Are you seeing a possibility that these idle mysql connections are not due to unused open persistent connections in the FPM pools? Given the above, it seems only natural that the behavior we are seeing is expected (though wouldn't always be present, depending on a number of factors, such as amount of traffic, ASG scaling logic, etc).

Based off what I think we can already deduce logically (essentially we should expect that `Rds Connections = ( ECS Web Tasks Count * 4 ) + ( Cavalcade ECS task Count * 4 )` (I think 4 cavalcade workers is the max?) ) then there is a maximum connections our RDS instances that we know. This is indeed complicated by the fact we have writers and readers on MySQL, and we can't predict exactly the distribution of persistent connections to each. Ideally we'd see a 50-50 split of connections to the read-replica gets half. That means the RDS Max Connections per instance could be 0.5x the total, but that's not guarenteed, so I'd imagine we actually want more like 70% Max connections capacity on the writer to account for an amount of "required mysql writer" requests.

_Aside: Currently with how LudicrousDB works with persistent connections, I don't know for sure if it is smart enough to detect an already open persistent connection, instead of randomly selecting either the reader or writer. This could mean more persistent connections are being held by PHP-FPM that strictly necessary_.

If this experiment is to evaluate the usefulness of persistent connections, from experience of why we added this, I can say it absolutely is necessary, as the connection time is very large, that it becomes impractically slow.

Given all the above, I don't really see anything that needs to be "fixed" (given all above points are correct). I do think we might need to look at an architectural / design change to reduce that count, but the reality seems to be: in a high traffic time, we'd expect to have `ECS Task Count * 4` workers (and therefore, necessarilya rds connections) which doesn't seem different to the reality we are observing on snopes.

nathanielks:
I now realize I based this assumption off a faulty calculation. I was comparing two similar environments and saw the second environment had a much lower connection count than the other. I didn't factor that the second environment had a second reader, so it distributed the connections across both of those, but I only calculated based off of one. Succintly, it does seem `Rds Connections = ( ECS Web Tasks Count * 4 ) + ( Cavalcade ECS task Count * 4 )` is roughly accurate:

Screen Shot 2020-06-26 at 9 56 04 AM

I'm in agreement that nothing is actually wrong and we just need to scale out when these connections get to the level we're seeing. I'll move this elsewhere!

PR #204, opened 2020-07-01
Backport v3: Catch exceptions from StopQuery in CloudWatch

joehoyle:
In cases where StopQuery fails, we don't want the exception to bubble, as it is non-essential.

See #202


roborourke:
Released in 3.0.12

PR #214, opened 2020-07-10
Cloud integration for Elasticsearch Service packages

roborourke:
Related to user dictionary support https://github.com/humanmade/altis-enhanced-search/pull/76

This updates provides the hooks into the search module packages functionality to interact with the AES API.


inc/aes_packages/namespace.php
```diff
@@ -0,0 +1,318 @@
+<?php
+/**
+ * AES Package File Integration.
```
joehoyle, line 3: AES seems to be an invented acronym? Assuming this is shows for AWS ElasticSearch? Could we call it `AWS_ES_Packages` at least?

inc/aes_packages/namespace.php
```diff
@@ -0,0 +1,318 @@
...
+use Exception;
+
+/**
+ * Hook into search module package support.
+ *
+ * @return void
+ */
+function setup() {
+	// Check the current environment and bail if it isn't AWS.
+	if ( ! in_array( Altis\get_environment_type(), [ 'development', 'staging', 'production' ], true ) ) {
```
joehoyle, line 21: This should use environment architecture, not type

inc/aes_packages/namespace.php
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Get Elasticsearch Service Client.
+ *
+ * @return ElasticsearchServiceClient
+ */
+function get_aes_client() : ElasticsearchServiceClient {
+	return Altis\get_aws_sdk()->createElasticsearchService( [
+		'version' => '2015-01-01',
+		'region' => Altis\get_environment_region(),
```
joehoyle, line 56: Region isn't needed here as it's set on the SDK IIRC.

inc/aes_packages/namespace.php
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * Create the package on AES and return the package ID string.
+ *
+ * @param string|null $package_id The package file path for ES mappings.
+ * @param string $slug The package slug.
+ * @param string $file The package file path.
+ * @param bool $for_network Whether this is a network level package or not.
+ * @return string|null
+ */
+function create_package_id( ?string $package_id, string $slug, string $file, bool $for_network = false ) : ?string {
```
joehoyle, line 76: IMO it would be better to allow `WP_Error` as a return, so the errors can be surfaced up the stack?

inc/aes_packages/namespace.php
```diff
@@ -0,0 +1,318 @@
...
+		$status = get_site_option( "altis_search_package_status_{$slug}", 'ASSOCIATING' );
+	} else {
+		$status = get_option( "altis_search_package_status_{$slug}", 'ASSOCIATING' );
+	}
+
+	// Queue up a background task if this is still pending.
+	// Association usually takes 1-5 minutes so use a 30 second resolution.
+	$hook_args = [ $package_id, $slug, $for_network ];
+	if ( $package_id && $status !== 'ACTIVE' && ! wp_next_scheduled( 'altis.search.check_package_status', $hook_args ) ) {
+		wp_schedule_single_event( time() + 60, 'altis.search.check_package_status', $hook_args );
```
joehoyle, line 185: This seems like an odd side-effect for a `get_` function to do, could this be moved to the caller?

inc/aes_packages/namespace.php
rmccue: Yeah, let's avoid AES, given it's a common acronym for encryption.

roborourke: Ah maybe that's where I picked it up. Sure I'd seen AES meaning Amazon Elasticsearch Service somewhere 🤷

inc/aes_packages/namespace.php
roborourke: Not easily. It's a bit of an odd one. This means it'll only do the lookup when the data is requested but I'll see if I can make the chain of background tasks more sensible.

inc/aes_packages/namespace.php
roborourke: Yeah the error handling sucks right now, I need to do some work on that part of it to better surface statuses and things in the UI.

inc/namespace.php
```diff
@@ -501,6 +501,11 @@ function load_plugins() {
 	if ( $config['healthcheck'] ) {
 		add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_healthcheck' );
 	}
+
+	// Bootstrap integration with Elasticsearch Service API.
+	if ( Altis\get_config()['modules']['search']['enabled'] ?? false ) {
```
joehoyle, line 506:
```suggestion
	if ( ! empty( Altis\get_config()['modules']['search']['enabled'] ) ) {
```

IMO this is more readable / simpler than the null-coalesce.

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/**
+ * Check and update the package association status.
+ *
+ * @param string $package_id The package ID to look up.
+ * @param string $slug The package slug.
+ * @param boolean $for_network Whether this is a network level package.
+ * @return void
+ * @throws Exception Thrown if there is any issue contacting the AES API, exceptions are caught.
```
joehoyle, line 205: It doesn't look this like this throw?

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,342 @@
...
+ * @param string $file The package file path.
+ * @param bool $for_network Whether this is a network level package or not.
+ * @return string|null
+ */
+function create_package_id( ?string $package_id, string $slug, string $file, bool $for_network = false ) : ?string {
+
+	// Check file is an S3 file path.
+	if ( strpos( $file, 's3://' ) !== 0 ) {
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		trigger_error( sprintf( 'The given package filepath %s is not a valid S3 file path.', $file ), E_USER_WARNING );
```
joehoyle, line 87: Sorry can't remember if I commented this before! For all these functions, can they return `WP_Error` so a message can be presented to the user, or is it not in a context where we could surface that? Typically `trigger_error` can be difficult to surface anywhere, especially for users.

inc/namespace.php
roborourke: I've got it in my head that it's safer to do it using null coalesce because `empty()` doesn't check each level of the array - I'm conflating it with `isset()`!

inc/elasticsearch_packages/namespace.php
roborourke: No I think this was added because PHPCS was being a bit dumb

inc/elasticsearch_packages/namespace.php
roborourke: Update this function to deal with WP_Errors. It's still a bit gnarly in the UI template but functional at least. Also added to the stored error reporting.

Think realistically I'd need to add an event log as things happen over time but not sure how far to take that really - I can display the package status to end users so that might be ok from a user perspective, any errors will be in the error logs so is that sufficient? Gonna do some more testing with a live instance again tomorrow.

joehoyle (approved):

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+}
+
+/**
+ * Filter the cron schedules list.
+ *
+ * @param array $schedules The list of available cron schedules.
+ * @return array
+ */
+function cron_schedules( array $schedules ) : array {
+	$schedules['1minute'] = [
```
rmccue, line 53: I thought we already had a `minutely` registered? We should probably add these to altis-core anyway really.

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	return Altis\get_aws_sdk()->createElasticsearchService( [
+		'version' => '2015-01-01',
+	] );
+}
+
+/**
+ * Return the domain ID portion of the Elasticsearch Service host name.
+ */
+function get_elasticsearch_domain() : string {
+	return preg_replace( '#^search-([a-z][a-z0-9\-]+)-[a-z0-9]+\..*$#', '$1', ELASTICSEARCH_HOST );
```
rmccue, line 89: I think this is more idiomatic modern PHP, as it doesn't overlap with PHP's interpolated variable string syntax:

```suggestion
	return preg_replace( '#^search-([a-z][a-z0-9\-]+)-[a-z0-9]+\..*$#', '\\1', ELASTICSEARCH_HOST );
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+ * Create the package on AES and return the package ID string.
+ *
+ * @param string|null $package_id The package file path for ES mappings.
+ * @param string $slug The package slug.
+ * @param string $file The package file path.
+ * @param bool $for_network Whether this is a network level package or not.
+ * @return string|WP_Error
+ */
+function create_package_id( ?string $package_id, string $slug, string $file, bool $for_network = false ) {
+
```
rmccue, line 102: Bonus newline.

```suggestion
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	// Check file is an S3 file path.
+	if ( strpos( $file, 's3://' ) !== 0 ) {
```
rmccue, line 104: Does this still apply locally?

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+
+	// Get Domain ID from host.
+	$domain = get_elasticsearch_domain();
+
+	// Get a unique package name.
+	// Must be no more than 28 chars.
+	$name = substr( sprintf(
+		'p%s-%s',
+		hash( 'crc32', $domain . $file . time() ),
+		trim( basename( $file, '.txt' ), '-' )
```
rmccue, line 125: Should this be sanitised?

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	), 0, 28 );
+
+	try {
+		// Derive S3 bucket and path.
+		preg_match( '#^s3://([^/]+)/(.*?)$#', $file, $s3_file_parts );
```
rmccue, line 130: It's possible for this to not match but still pass the above check for the `s3://` prefix; e.g. `s3://bukkit` will pass the above sanity check, but `preg_match` will return false here.

Suggest checking the return value to ensure it actually matches.

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	if ( in_array( $status, [ 'AVAILABLE', 'COPYING', 'VALIDATING' ], true ) ) {
+		wp_schedule_event( time(), '1minute', 'altis.search.check_package_status', [
+			$package_id,
+			$slug,
+			$for_network,
+			$existing_package_id,
+		] );
+	} else {
+		return new WP_Error( 'create_package_error', sprintf( 'Package upload error, current status is %s', $status ) );
+	}
```
rmccue, line 179: This conditional should be flipped to return early and remove the else branch.
```suggestion
	if ( ! in_array( $status, [ 'AVAILABLE', 'COPYING', 'VALIDATING' ], true ) ) {
		return new WP_Error( 'create_package_error', sprintf( 'Package upload error, current status is %s', $status ) );
	}

	wp_schedule_event( time(), '1minute', 'altis.search.check_package_status', [
		$package_id,
		$slug,
		$for_network,
		$existing_package_id,
	] );
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+/**
+ * Check whether the package has finished associating with the AES domain
+ * before returning it.
+ *
+ * @param string|null $package_id The stored package ID.
+ * @param string $slug The slug for the package.
+ * @param boolean $for_network Whether this is a network level package.
+ * @return string|null
+ */
+function get_package_id( ?string $package_id, string $slug, bool $for_network = false ) {
```
rmccue, line 194:
```suggestion
function get_package_id( ?string $package_id, string $slug, bool $for_network = false ) : ?string {
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	return $package_id;
+}
+
+/**
+ * Add a package ID to the list of packages to be removed.
+ *
+ * @param string $package_id The package ID to remove.
+ * @return void
+ */
+function add_package_to_remove( string $package_id ) {
```
rmccue, line 215:
```suggestion
function add_package_to_remove( string $package_id ) : void {
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+		return new WP_Error( 'create_package_error', sprintf( 'Package upload error, current status is %s', $status ) );
+	}
+
+	// Return the AES package ID string.
+	return $package_id;
+}
+
+/**
+ * Check whether the package has finished associating with the AES domain
+ * before returning it.
```
rmccue, line 187: This doesn't seem correct?

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+ * Check and update the package association status.
+ *
+ * @param string $package_id The package ID to look up.
+ * @param string $slug The package slug.
+ * @param boolean $for_network Whether this is a network level package.
+ * @param string|null $existing_package_id An existing package ID to be removed.
+ * @return void
+ */
+function on_check_package_status( string $package_id, string $slug, bool $for_network = false, ?string $existing_package_id = null ) {
+
```
rmccue, line 241: Bonus newline.

```suggestion
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+		if ( $for_network ) {
+			update_site_option( "altis_search_package_status_{$slug}", $status );
+			update_site_option( "altis_search_package_error_{$slug}", $package['ErrorDetails']['ErrorMessage'] ?? null );
+		} else {
+			update_option( "altis_search_package_status_{$slug}", $status );
+			update_option( "altis_search_package_error_{$slug}", $package['ErrorDetails']['ErrorMessage'] ?? null );
+		}
+
+		// If the package is now active then update the index settings.
+		if ( $status === 'ACTIVE' ) {
```
rmccue, line 301: Ideally these conditions should be switched so that the "happy path" is the main flow through the code (and so that the conditional itself is shorter)

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	}
+}
+
+/**
+ * Loop over packages to remove and delete them. Called on
+ * the altis.search.updated_all_index_settings hook.
+ *
+ * @return void
+ */
+function on_update_index_settings() {
```
rmccue, line 347:
```suggestion
function on_update_index_settings() : void {
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	// Empty the array of packages to be removed.
+	update_site_option( 'altis_search_packages_to_remove', [] );
+}
+
+/**
+ * Hook triggered when deleting a package directly.
+ *
+ * @param string $package_id The old package ID that has been deleted.
+ */
+function on_deleted_package( string $package_id ) {
```
rmccue, line 363:
```suggestion
function on_deleted_package( string $package_id ) : void {
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	// Empty the array of packages to be removed.
+	update_site_option( 'altis_search_packages_to_remove', [] );
+}
+
+/**
+ * Hook triggered when deleting a package directly.
+ *
+ * @param string $package_id The old package ID that has been deleted.
```
rmccue, line 361:
```suggestion
 * @param string $package_id The old package ID that has been deleted.
 * @return void
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+	foreach ( $packages_to_remove as $package_id ) {
+		delete_package( $package_id );
+	}
+
+	// Empty the array of packages to be removed.
+	update_site_option( 'altis_search_packages_to_remove', [] );
+}
+
+/**
+ * Hook triggered when deleting a package directly.
```
rmccue, line 359: Switch to imperative(?):
```suggestion
 * Handle direct package deletion.
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+
+/**
+ * Delete a package off Elasticsearch Service.
+ *
+ * @param string $package_id The package ID to delete.
+ * @param int $max_retries The maximum number of times to attempt deletion.
+ * @return void
+ */
+function delete_package( string $package_id, int $max_retries = 5 ) {
+
```
rmccue, line 375:
```suggestion
function delete_package( string $package_id, int $max_retries = 5 ) : void {
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+				] );
+			}
+
+			// Reschedule deletion attempt.
+			wp_schedule_single_event( time() + ( 5 * MINUTE_IN_SECONDS ), 'altis.search.delete_package', [
+				$package_id,
+				$max_retries - 1,
+			] );
+
+			return;
```
rmccue, line 412: Should we return a code to indicate to the consumer whether deletion occurred vs was scheduled vs failed?

inc/namespace.php
```diff
@@ -515,6 +515,11 @@ function load_plugins() {
 	if ( $config['healthcheck'] ) {
 		add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_healthcheck' );
 	}
+
+	// Bootstrap integration with Elasticsearch Service API.
+	if ( ! empty( Altis\get_config()['modules']['search']['enabled'] ) ) {
+		Elasticsearch_Packages\setup();
```
rmccue, line 521:
```suggestion
		Elasticsearch_Packages\bootstrap();
```

rmccue (changes requested):

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,432 @@
...
+use WP_Error;
+
+/**
+ * Hook into search module package support.
+ *
+ * @return void
+ */
+function setup() {
+	// Check the current environment and bail if it isn't AWS.
+	if ( ! in_array( Altis\get_environment_architecture(), [ 'ec2', 'ecs' ], true ) ) {
```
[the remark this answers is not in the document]
joehoyle: yeah that would be good

inc/elasticsearch_packages/namespace.php
roborourke: Not sure, I feel like it needs to be here to be completely reliable.

rmccue: If altis/cloud depends on altis/core, that'd be reliable? IMO, our modules should all depend on core

roborourke: There's no `minutely` schedule registered in any of the Altis modules themselves. I'll update this to be called `minutely` anyway so we could add some additional schedules in the CMS or core module in future

inc/elasticsearch_packages/namespace.php
roborourke: From the php.net docs: https://www.php.net/manual/en/function.preg-replace.php

> replacement may contain references of the form \n or $n, with the latter form being the preferred one.

inc/elasticsearch_packages/namespace.php
roborourke: This whole cloud integration is ignored on local envs.

inc/elasticsearch_packages/namespace.php
roborourke: Maybe, but AWS will throw an error if the string is invalid anyway.

inc/elasticsearch_packages/namespace.php
roborourke: Not sure where that would be used / useful right now. This is a background task.

inc/elasticsearch_packages/namespace.php
roborourke: I guess that's describing what this filter callback is doing eg. a long description. 
```suggestion
 * Get the package ID.
 *
 * This callback checks whether the package has finished associating with the AES domain
 * before returning its ID.
```

inc/elasticsearch_packages/namespace.php
rmccue: 👍 I guess it's just a common practice we have. Should def add to altis-core in future, I'll file an issue.

inc/elasticsearch_packages/namespace.php
rmccue: Classic PHP. Welp, fine by me then!

inc/elasticsearch_packages/namespace.php
rmccue: That's true, but the function itself (`delete_package()`) seems like a generic API that anyone could use internally?

roborourke: Yeah that's true. Makes me wonder if there's a nicer way async functions like this could be packaged. Feels quite separate from the action hooks required to make it work.

rmccue (approved):

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,447 @@
...
+ * @return ElasticsearchServiceClient
+ */
+function get_es_client() : ElasticsearchServiceClient {
+	return Altis\get_aws_sdk()->createElasticsearchService( [
+		'version' => '2015-01-01',
+	] );
+}
+
+/**
+ * Return the domain ID portion of the Elasticsearch Service host name.
```
rmccue, line 90:
```suggestion
 * Get the domain ID portion of the Elasticsearch Service host name.
 *
 * @return string
```

inc/elasticsearch_packages/namespace.php
```diff
@@ -0,0 +1,447 @@
...
+/**
+ * Check and update the package association status.
+ *
+ * @param string $package_id The package ID to look up.
+ * @param string $slug The package slug.
+ * @param boolean $for_network Whether this is a network level package.
+ * @param string|null $existing_package_id An existing package ID to be removed.
+ * @return void
+ */
+function on_check_package_status( string $package_id, string $slug, bool $for_network = false, ?string $existing_package_id = null ) {
```
rmccue, line 250:
```suggestion
function on_check_package_status( string $package_id, string $slug, bool $for_network = false, ?string $existing_package_id = null ) : void {
```

rmccue (approved):
Approved; two minor nits in suggestions, then can merge at will.

PR #215, opened 2020-07-15
Check architecture when adding ec2 metadata

roborourke:
Fetching EC2 instance metadata spits out an error on local environments. This was never spotted until #210 got merged and fixed the status check.


joehoyle:
Would like to actually only have the cloud module on cloud at some point!

roborourke:
🤔 yeah, that makes sense given its purpose. Will have to see what components of local-server and local-chassis are using it. As long as we can avoid duplicating any code between this module and those.

PR #218, opened 2020-07-20
Ensure Tachyon and Uploads URL use current site host name

### LINKED ISSUE #217: Uploads and Tachyon URLs should match site domain name
Sites are set up initially on an `*.altis.cloud` domain on our infrastructure, this is the primary host name set for the site. When the real production domain names are pointed to the site (or subsite) the uploads URL is still hardcoded to the original `*.altis.cloud` URL via the `S3_UPLOADS_BUCKET_URL` constant.

The same is true for Tachyon URLs. Both of these URLs should match the URL defined in the database for the site.

These URLs can both be filtered and should be handled by default in this module.

Acceptance criteria:

- [ ] Tachyon URL should match the host name of the current site
- [ ] Uploads URL should match the host name of the current site
- [ ] In both cases if it is a subsite using a subfolder path don't change anything

roborourke:
Currently we define the S3 Uploads bucket URL at the infrastructure level, eg. `example.altis.cloud/uploads`.

Once the site has been migrated to its production URL the infra level setting is still used. Ideally it should change to match the URL set in the database.

This update ensures the S3 bucket URL and Tachyon URL use the main site host name for all sub domain and sub directory sites, and if a subsite has a custom domain ensures it uses that.

Fixes #217

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -484,12 +484,12 @@ function load_plugins() {
 	// at the infra level currently.
 	if ( ! defined( 'TACHYON_URL' ) ) {
 		// Override the default host name for Tachyon to match the current site.
-		add_filter( 'tachyon_url', __NAMESPACE__ . '\\set_tachyon_hostname' );
+		add_filter( 'tachyon_url', __NAMESPACE__ . '\\set_tachyon_hostname', 11 );
 		define( 'TACHYON_URL', get_site_url( get_main_site_id( get_main_network_id() ), '/tachyon' ) );
 	}
 
 	if ( $config['s3-uploads'] ) {
-		add_filter( 'upload_dir', __NAMESPACE__ . '\\set_s3_uploads_bucket_url_hostname' );
+		add_filter( 'upload_dir', __NAMESPACE__ . '\\set_s3_uploads_bucket_url_hostname', 11 );
 		require_once Altis\ROOT_DIR . '/vendor/humanmade/s3-uploads/s3-uploads.php';
 	}
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -539,9 +539,6 @@ function set_s3_uploads_bucket_url_hostname( array $dirs ) : array {
 	$current_host = wp_parse_url( site_url(), PHP_URL_HOST );
 
 	// Ensure uploads host at least matches primary site host.
-	error_log( $primary_host );
-	error_log( $s3_host );
-	error_log( $current_host );
 	if ( strpos( $primary_host, $s3_host ) === false ) {
 		$dirs['url'] = str_replace( $s3_host, $primary_host, $dirs['url'] );
 		$dirs['baseurl'] = str_replace( $s3_host, $primary_host, $dirs['baseurl'] );
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -465,6 +465,16 @@ function load_db() {
 }
 
 /**
+ * Return the site URL for the main site on the network.
+ *
+ * @param string $path Optional path to append to URL.
+ * @return string
+ */
+function get_main_site_url( string $path = '' ) : string {
+	return get_site_url( get_main_site_id( get_main_network_id() ), $path );
+}
+
+/**
  * Load the plugins in altis.
  */
 function load_plugins() {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -503,6 +507,52 @@ function load_plugins() {
...
+/**
+ * Ensure the S3 Uploads Bucket URL matches the current site hostname.
+ *
+ * @param array $dirs Uploads directories array.
+ * @return array
+ */
+function set_s3_uploads_bucket_url_hostname( array $dirs ) : array {
+	$s3_uploads = S3_Uploads::get_instance();
+
+	$primary_host = wp_parse_url( get_site_url( get_main_site_id( get_main_network_id() ) ), PHP_URL_HOST );
```
rmccue, line 537: We're using this here and in `TACHYON_URL` above, can we abstract given it's a somewhat complex thing? Also should both have error handling?

inc/namespace.php
```diff
@@ -503,6 +507,52 @@ function load_plugins() {
...
+	$s3_host = wp_parse_url( $s3_uploads->get_s3_url(), PHP_URL_HOST );
+	$current_host = wp_parse_url( site_url(), PHP_URL_HOST );
+
+	// Ensure uploads host at least matches primary site host.
+	if ( strpos( $primary_host, $s3_host ) === false ) {
```
rmccue, line 542: Should check that it's at the end rather than just exists?

inc/namespace.php
```diff
@@ -503,6 +517,83 @@ function load_plugins() {
...
+		return $tachyon_url;
+	}
+
+	if ( ! $current_host ) {
+		trigger_error( sprintf( 'Error parsing current site URL: %s', esc_url_raw( site_url() ) ), E_USER_WARNING );
+		return $tachyon_url;
+	}
+
+	// Only do the replacement if the host name is not a subdomain of the Tachyon host.
+	if ( substr( $current_host, -strlen( $tachyon_host ) ) !== $tachyon_host ) {
```
rmccue, line 541: Can we make this `-1 * strlen ...` instead? It's clearer than an easily missed symbol and more apparent that you're applying a mathematical operator rather than a weird invocation of a function.
```suggestion
	if ( substr( $current_host, -1 * strlen( $tachyon_host ) ) !== $tachyon_host ) {
```

inc/namespace.php
```diff
@@ -503,6 +517,83 @@ function load_plugins() {
...
+		return str_replace( $tachyon_host, $current_host, $tachyon_url );
```
rmccue, line 542: Can we anchor this with `://` as well to ensure that we're not accidentally replacing in filenames?

inc/namespace.php
```diff
@@ -503,6 +517,83 @@ function load_plugins() {
...
+	// Ensure uploads host at least matches primary site host.
+	if ( $s3_host !== $primary_host ) {
+		$dirs['url'] = str_replace( $s3_host, $primary_host, $dirs['url'] );
+		$dirs['baseurl'] = str_replace( $s3_host, $primary_host, $dirs['baseurl'] );
+	}
+
+	// Only do the replacement if the host name is not a subdomain of the S3 host.
+	if ( substr( $current_host, -strlen( $primary_host ) ) !== $primary_host ) {
+		$dirs['url'] = str_replace( $primary_host, $current_host, $dirs['url'] );
+		$dirs['baseurl'] = str_replace( $primary_host, $current_host, $dirs['baseurl'] );
```
rmccue, line 591: Ditto above comment for all of these hosts.

inc/namespace.php
```diff
@@ -503,6 +517,83 @@ function load_plugins() {
...
+	}
+
+	// Ensure uploads host at least matches primary site host.
+	if ( $s3_host !== $primary_host ) {
+		$dirs['url'] = str_replace( $s3_host, $primary_host, $dirs['url'] );
+		$dirs['baseurl'] = str_replace( $s3_host, $primary_host, $dirs['baseurl'] );
+	}
+
+	// Only do the replacement if the host name is not a subdomain of the S3 host.
+	if ( substr( $current_host, -strlen( $primary_host ) ) !== $primary_host ) {
```
rmccue, line 589: Ditto above operator comment.
```suggestion
	if ( substr( $current_host, -1 * strlen( $primary_host ) ) !== $primary_host ) {
```

inc/namespace.php
roborourke: That not part of the host name though is it?

roborourke: ohhh I getcha, yes

PR #226, opened 2020-08-04
Update build container package versions

nathanielks:
This brings our docs up to date with the latest updates.

### DIFF
--- a/docs/build-scripts.md
+++ b/docs/build-scripts.md
@@ -33,7 +33,7 @@ console.log( 'Starting build...' )
 The build script is run in an integration environment, not in the context of the application servers. The following packages are available to your script:
 
 - PHP 7.2
-- Composer
-- NodeJS 8.11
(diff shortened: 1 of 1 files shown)


nathanielks:
I think I got the labels right, feel free to edit as you please!

rmccue:
FWIW, v1 is unsupported, and we have no clients left on v2, so we generally don't bother backporting those. :)

nathanielks:
Hmm, I just checked the site and the old versions are still being displayed. Would you help with that, @rmccue ?

https://www.altis-dxp.com/resources/docs/cloud/build-scripts/

roborourke:
Just need to update the marketing site it seems. Looks like there might be other work under way on the dev branch though. It’ll be updated to v5 in a week anyway which def has the changes

nathanielks:
Bril, thanks @roborourke !

Issue #183, opened 2020-06-11
Add final filtered query into X-Ray


Nikschavan:
The query which is built from the application code is being tracked to X-Ray - 

https://github.com/humanmade/altis-cloud/blob/a8c4eb12098b344df9843e8c51e4806935653a35/inc/class-db.php#L65

When this query is being executed in `WPDB::query` it is passed through the filter [`query`](https://github.com/WordPress/WordPress/blob/8d1e51e9c769ec57375efb39b72cac2c3abf8108/wp-includes/wp-db.php#L1894) which removes placeholders at the very least and it is also possible that queries can be modified altogether using this filter.

Since when sending the query to X-Ray does not take into account the filter we don't get the actual query that was executed. 

Running filter `query` just before sending it to X-Ray should hopefully fix this.

roborourke:
This might be intentional as a means of making sure the query is redacted as it could potentially show email addresses and such. We already redact some data before sending it to XRay so we need to make the same consideration here, I'm not sure it's as simple as changing the hook.

Nikschavan:
I only noticed one of the PRs with redactions in X-Ray - https://github.com/humanmade/aws-xray/pull/34

If we run this filter, [WordPress's functions to replace the placeholders](https://github.com/WordPress/WordPress/blob/32edd58e4c8e42b68813e3b92886aaa3e50cc1a9/wp-includes/wp-db.php#L2104) gets executed and removes the placeholders.

I did not find any query redactions atm, Can you point me to ones I am missing so that I can check if these are working as expected with this change?

rmccue:
These escapes aren't redactions, they're just placeholders for the `%` character so it survives any `sprintf` calls without being mangled. We can safely change those back.

jennybeaumont:
This was completed

PR #235, opened 2020-08-09
Send final filtered database query to X-Ray

Nikschavan:
See #183


Nikschavan:
Similar to https://github.com/humanmade/hm-platform/pull/184

Nikschavan:
@joehoyle this is ready for a review.

Not exactly sure if we still need to override the method `query()` after moving the `trace_wpdb_query()` call to `_do_query()`?

joehoyle (approved):
MM think this is good, we can keep the QM integration in `query`.

PR #239, opened 2020-08-26
Update default ignored cache query vars

roborourke:
This adds some extra default ignored query string vars to help improve page caching.


joehoyle:
I think the Page Cache docs will need updating here too.

roborourke:
Thanks @joehoyle, good catch - updated these now

PR #242, opened 2020-08-27
Fix InvalidSequenceToken in CloudWatch Logs

### LINKED ISSUE #168: Cloudwatch HTTP error: invalid sequence token
When running a command on a client's sandbox we get the following Warning:
```
Warning: Error executing "PutLogEvents" on "https://logs.us-east-1.amazonaws.com"; AWS HTTP error: Client error: `POST https://logs.us-east-1.amazonaws.com` resulted in a `400 Bad Request` response:
{"__type":"InvalidSequenceTokenException","expectedSequenceToken":"49605576650224253646933900996881587352498596412622676 (truncated...)
 InvalidSequenceTokenException (client): The given sequenceToken is invalid. The next expected sequenceToken is: 49605576650224253646933900996881587352498596412622676434 - {"__type":"InvalidSequenceTokenException","expectedSequenceToken":"49605576650224253646933900996881587352498596412622676434","message":"The given sequenceToken is invalid. The next expected sequenceToken is: 49605576650224253646933900996881587352498596412622676434"} in /usr/src/app/vendor/altis/cloud/inc/cloudwatch_logs/namespace.php on line 54```

### PR DESCRIPTION
**Pull request by joehoyle**
In cases when the sequence token in the cache is wrong, we can auto-retry with the new sequence token.

I also added unit tests here, which I'm running from the main platform-dev repo. I will share my approach on that, but in the mean time I don't see an issue per-say with including the unit tests already.

Fixes #168.

### DIFF
--- a/tests/class-test-cloudwatch-logs.php
+++ b/tests/class-test-cloudwatch-logs.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Altis\Cloud\Tests;
+
+use Altis\Cloud\CloudWatch_Logs;
+use Aws\CommandInterface;
+use Aws\Exception\AwsException;
+use Aws\MockHandler;
+use Aws\Result;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\RequestInterface;
+
+class Test_CloudWatch_Logs extends TestCase {
+	/**
+	 * SDK mock handler.
+	 *
+	 * @see https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_handlers-and-middleware.html
+	 *
+	 * @var MockHandler
+	 */
+	protected $sdk_mock;
+
+	public function setUp() {
+		$this->sdk_mock = new MockHandler();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/class-test-cloudwatch-logs.php [hunk 1]
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Altis\Cloud\Tests;
```
joehoyle, line 3: Sorry what?

roborourke: Weird. I think you just add `phpcs:disable HM.Files.NamespaceDirectoryName.NameMismatch` for now. Seems like maybe a bug.

roborourke (approved):

tests/class-test-cloudwatch-logs.php
rmccue: I think this is from an old version of the coding standards that did some `substr()` wrong, I think we fixed this?

rmccue (approved):

rmccue:
Do we know what causes these in the first place btw?

joehoyle:
Pretty sure it's due to races between processes.

PR #243, opened 2020-08-27
Exclude ALB healthchecks from X-ray

### LINKED ISSUE #237: Exclude healthchecks from X-Ray
Healthcheck requests shouldn't be loaded in X-Ray, as they affect the statistics of request times, etc.

joehoyle:
Fixes #237

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -44,11 +44,14 @@ const WILDCARD_INVALIDATION_LIMIT = 10;
 function bootstrap() {
 	$config = get_config();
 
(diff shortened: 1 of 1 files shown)


rmccue (approved):
Wondering if we should keep the data but change where it's recorded for our own internal debugging purposes. Seems like it might be useful occasionally.

joehoyle:
The best option I think there would be to tag with another service name instead of the main app one, like `humanade-production-healthcheck`. However, I'm not sure it's worth it, I know it _can_ be useful once in a blue moon.

rmccue:
Right, that's what I mean. We can handle in the future; if you think it's worth having, can you create a new issue to track that please?

joehoyle:
Ok created in https://github.com/humanmade/altis-cloud/issues/246

PR #275, opened 2020-11-11
Fix object cache method check

### LINKED ISSUE #263: Incorrect check against object cache for Redis in health check
https://github.com/humanmade/altis-cloud/blob/3c7c3d33453b86cefc7b2aaae80b0c4f96e94306/inc/healthcheck/namespace.php#L165

roborourke:
Fixes #263

### DIFF
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -43,7 +43,6 @@ function bootstrap() {
  * @return void
  */
 function output_page( array $checks ) {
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):
This looks good to me, what's up with `getStatus`? It doesn't look like anything is actually replacing that method in this PR.

rmccue:
I think it was a misspelling of `getStats` in the first one and a copypasta in the second one.

rmccue (approved):

PR #276, opened 2020-11-11
Update dashboard logs images

### LINKED ISSUE #257: Localise images linked from S3
In `logs.md` we have some images that are linked from S3. These should be moved into the docs/assets directory and the URLs updated. In addition they should have alt text added.

roborourke:
Also localises the images rather than using the S3 bucket.

Fixes #257


jazzsequence (approved):
👍

PR #301, opened 2020-12-17
Add Fluent Bit logging with CloudWatch fallback

nathanielks:
There have been a number of cases where, when a site is under load, errors are thrown due to writing logs directly to CloudWatch's API. This is bad because the CloudWatch API may not be available or it's slow to respond. We are updating the infrastructure to expose a Fluent Bit logging daemon which can be used to forward these logs in a more performant manner.

See https://github.com/humanmade/terraform-app-stack/issues/715

Acceptance Criteria:
- [x] Current logger works when Fluent Bit is not available for the environment
- [x] Fluent Bit is used when it is available


inc/cavalcade_runner_to_cloudwatch/namespace.php
```diff
@@ -102,25 +103,17 @@ function on_end_job( Worker $worker, Job $job, string $status ) {
 	$error_output_property->setAccessible( true );
 	$error_output = $error_output_property->getValue( $worker );
 
-	send_event_to_stream(
```
nathanielks, line 105: Is there a reason we were using this copied function instead of the `send_events_to_stream` defined in `inc/cloudwatch_logs`?

inc/fluent_bit/error_handler/namespace.php
```diff
@@ -0,0 +1,112 @@
...
+
+	$logger = Fluent_Bit\get_logger( 'app.php-structured.' . $error['type'] );
+	$logger->error( $json );
+
+	// TODO is this necessary to keep?
+	$altis_cloudwatch_error_handler_errors[ $errno ][] = [
+		'timestamp' => time() * 1000,
+		'message'   => $json, // @codingStandardsIgnoreLine
+	];
+	$altis_cloudwatch_error_handler_error_count++;
```
nathanielks, line 66: With Fluent Bit, we don't need to send logs in batches. Is this necessary anymore?

inc/fluent_bit/namespace.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Retrieve logger for specied $tag_name
+ *
+ * @param string $tag_name Name of the tag to route log messages to
+ * @return Monolog\Logger
+ */
+function get_logger( string $tag_name ) {
+	// TODO how do I need to name the file/class to be autoloaded?
+	require_once __DIR__ . '/class-msgpackformatter.php';
```
nathanielks, line 31: I've had trouble getting autoloading to work, does this look right?

inc/namespace.php
```diff
@@ -46,6 +48,10 @@ function bootstrap() {
 
 	$is_alb_healthcheck = isset( $_SERVER['HTTP_USER_AGENT'] ) && strpos( $_SERVER['HTTP_USER_AGENT'], 'ELB-HealthChecker' ) === 0;
 
+	require_once __DIR__ . '/fluent_bit/namespace.php';
+	require_once __DIR__ . '/fluent_bit/error_handler/namespace.php';
+	Fluent_Bit\Error_Handler\bootstrap();
+
```
nathanielks, line 54: Do we want to leave this as is or have it enabled behind a feature flag?

nathanielks:
Okie doke, I feel this is ready for a cursory review. This is designed to be backwards compatible so it can be deployed before or after Fluent Bit has been added to the infrastructure. For instance, a new logging function has been introduced:

```
$result = log_to_cloud( 'audit-log', 'items', json_encode( $data ) )
```

If Fluent Bit is available, it will send logs there, otherwise use the existing CloudWatch logger.

So far I've been manually setting the required constants (FLUENT_HOST, FLUENT_PORT) and using a Fluent container outside of local server. Should I also add a fluent container to local-server as well?

inc/fluent_bit/error_handler/namespace.php
```diff
@@ -0,0 +1,112 @@
+<?php
```
joehoyle, line 1: Could we move this out of the fluent_bit namespace, and then provide Cloud APIs that are fluent agnostic, thinking that would be better decoupling. I should have do that with the CloudWatch logs too probably!

inc/fluent_bit/namespace.php
```diff
@@ -0,0 +1,55 @@
...
+use Altis\Cloud\Fluent_Bit\MsgPackFormatter;
+use Monolog\Handler\SocketHandler;
+use Monolog\Logger;
+
+/**
+ * Check if required constants have been defined
+ *
+ * @return boolean
+ */
+function available() {
```
joehoyle, line 19: `is_*` is typical for Boolean functions

inc/fluent_bit/namespace.php
joehoyle, line 31: Typically just include it at the top level, I think the main cloud module file has all the includes IIRC.

inc/fluent_bit/namespace.php
```diff
@@ -0,0 +1,55 @@
...
+
+	$logger = new Logger( $tag_name );
+
+	// Use Fluent Bit if it's available.
+	if ( available() ) {
+		$socket = new SocketHandler( FLUENT_HOST . ':' . FLUENT_PORT, Logger::DEBUG );
+		$socket->setFormatter( new MsgPackFormatter() );
+		$logger->pushHandler( $socket );
+	} else {
+		trigger_error( 'Fluent Bit is not available. Logs will not be routed anywhere.', E_USER_WARNING );
```
joehoyle, line 50: Should we just fall back to error_log maybe?

inc/namespace.php
```diff
@@ -954,3 +960,43 @@ function set_wp_debug_constants() : void {
...
+	}
+
+	// Basic error correction to ensure $level is an allowed value.
+	if ( ! in_array( $level, [ 'debug', 'info', 'notice', 'warning', 'error', 'critical', 'alert', 'emergency' ], true ) ) {
+		$level = 'info';
+	}
+
+	$fluent_tag = sprintf( 'app.%s.%s', $log_group, $log_stream );
+	$logger = Fluent_Bit\get_logger( $fluent_tag );
+	$logger->$level( $message );
```
joehoyle, line 999: Is there a generic `log` function that takes the level as a param? thinking that might be cleaner than a dynamically called method.

inc/ses_to_cloudwatch/namespace.php
```diff
@@ -7,8 +7,8 @@
 
 namespace Altis\Cloud\SES_To_CloudWatch;
 
-use Altis\Cloud\CloudWatch_Logs;
 use Exception;
+use function Altis\Cloud\log_to_cloud;
```
joehoyle, line 11: I think we disallowed `use function` in recent coding standards. You should always `use Altis\Cloud` and then call `Cloud\log_to_cloud`

inc/namespace.php
```diff
@@ -954,3 +960,43 @@ function set_wp_debug_constants() : void {
...
+ *
+ * @param string $log_group Name of the log group to send logs to. Environment
+ *                          name is added automatically.
+ * @param string $log_stream Name of the log stream.
+ * @param string $message Message to log
+ * @param string $level Level of the log. Corresponds to Psr\Log\LoggerInterface
+ * @see https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#3-psrlogloggerinterface
+ * @return boolean
+ */
+function log_to_cloud( string $log_group, string $log_stream, string $message, string $level = 'info' ) : bool {
```
joehoyle, line 976: Wondering if there's a better function name here, @rmccue for ideas perhaps.

joehoyle:
Looking good!

inc/namespace.php
nathanielks: Bah, there totally is 🤦 Good call 👍

inc/fluent_bit/namespace.php
nathanielks: I like that better 👍

inc/namespace.php
nathanielks: Yeah, totally open to better ideas.

inc/fluent_bit/error_handler/namespace.php
nathanielks: Can do!

inc/fluent_bit/error_handler/namespace.php
nathanielks: @joehoyle do you know if this isn't needed anymore?

inc/namespace.php
nathanielks: Considering we disallow `use function`, it could simply be called `log()`? It'd be used as `Cloud\log()`.

inc/fluent_bit/namespace.php
roborourke: @nathanielks you would need to add a `classmap` entry to the `composer.json` specifying the directories to read classes from. See the composer docs here https://getcomposer.org/doc/04-schema.md#autoload

inc/namespace.php
roborourke: I think it's an order of priority thing - if Fluent it is available we should use it if not then fall back. While releases aren't tied to infrastructure versions I think we'll need to be backwards compatible.

Make this conditional based on whether `Altis\Cloud\Fluent_Bit\is_available()` is true I suppose. That check could be done inside the `Fluent_Bit\Error_Handler\bootstrap()` function I suppose. No harm loading the PHP files even if they're unused.

inc/namespace.php
```diff
@@ -46,6 +48,10 @@ function bootstrap() {
 
 	$is_alb_healthcheck = isset( $_SERVER['HTTP_USER_AGENT'] ) && strpos( $_SERVER['HTTP_USER_AGENT'], 'ELB-HealthChecker' ) === 0;
 
+	require_once __DIR__ . '/fluent_bit/namespace.php';
+	require_once __DIR__ . '/fluent_bit/error_handler/namespace.php';
```
roborourke, line 52: Include these via the `composer.json` autoload section. If you're having issues with the autoload.php not including them then try removing `composer.lock` and `vendor/altis/cloud` then running `composer update`.

inc/fluent_bit/namespace.php
nathanielks: Thank you, @roborourke !

inc/namespace.php
nathanielks: Aha! Thank for you noting how to update dependencies, that was helpful!

inc/namespace.php
nathanielks: We'll use `Cloud\get_logger()` instead.

inc/fluent_bit/namespace.php
nathanielks: I ended up ommitting this. It'll  fall back to logging to CLoudWatch directly if Fluent Bit isn't available.

nathanielks:
This is ready for final review!

roborourke:
Just some linter errors to fix up here 👍

nathanielks:
@roborourke fixed up!

inc/fluent_bit/class-msgpackformatter.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Monolog\Formatter\FluentdFormatter;
+use RuntimeException;
+
+/**
+ * Class MsgPackFormatter
+ *
+ * Serializes a log message to Fluent Bit socket protocol
+ *
+ * @author Nathaniel Schweinberg <nathaniel@humanmade.com>
```
rmccue, line 28: Individual credit not needed, work is copyright Human Made and handled at the package level.

```suggestion
```

inc/fluent_bit/class-msgpackformatter.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+class MsgPackFormatter extends FluentdFormatter {
+
```
rmccue, line 31: Bonus newline.

```suggestion
```

inc/fluent_bit/class-msgpackformatter.php
```diff
@@ -0,0 +1,78 @@
...
+	 * @throws RuntimeException Break when msgpack extension is not available
+	 */
+	public function __construct( bool $levelTag = false ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName
+		if ( ! function_exists( 'msgpack_pack' ) ) {
+			throw new RuntimeException( 'PHP\'s msgpack extension is required to use Monolog\'s MsgPackFormatter' );
+		}
+
+		// phpcs:ignore WordPress.NamingConventions.ValidVariableName
+		$this->levelTag = $levelTag;
+	}
```
rmccue, line 47: Doesn't seem like there's a good reason to ignore the coding standard for the parameter name. This _might_ now throw another error for the property name, but given that's inherited, that specific part can be ignored (I think it's a NamingConventions.ValidPropertyUse or something)

```suggestion
	 * @param boolean $level_tag whether to append log level to Fluent Bit tag
	 * name or not.
	 * @return self
	 * @throws RuntimeException Break when msgpack extension is not available
	 */
	public function __construct( bool $level_tag = false ) {
		if ( ! function_exists( 'msgpack_pack' ) ) {
			throw new RuntimeException( 'PHP\'s msgpack extension is required to use Monolog\'s MsgPackFormatter' );
		}

		$this->levelTag = $level_tag;
	}
```

inc/fluent_bit/class-msgpackformatter.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Normalizes and formats the log record using the msgpack serialization
+	 * format.
+	 *
+	 * @param array $record Monolog log record
+	 * @return string The formatted log record
+	 */
+	public function format( array $record ): string {
```
rmccue, line 56: I think this needs an extra space (although I guess we don't sniff for this)

```suggestion
	public function format( array $record ) : string {
```

inc/namespace.php
```diff
@@ -674,7 +680,7 @@ function reflect_cloudfront_headers() {
 	foreach ( $headers as $header ) {
 		$header_key = 'HTTP_' . str_replace( '-', '_', strtoupper( $header ) );
 		if ( isset( $_SERVER[ $header_key ] ) ) {
-			header( sprintf( 'X-%s: %s', $header, $_SERVER[ $header_key ] ) );
+			header( sprintf( 'X-%s: %s', $header, wp_unslash( $_SERVER[ $header_key ] ) ) );
```
rmccue, line 683: This sniff can safely be ignored as we're using a safe subset of the headers.
```suggestion
			// phpcs:ignore HM.Security.ValidatedSanitizedInput.InputNotSanitized -- Using safe subset of headers.
			header( sprintf( 'X-%s: %s', $header, wp_unslash( $_SERVER[ $header_key ] ) ) );
```

inc/namespace.php
```diff
@@ -854,6 +860,22 @@ function get_cloudfront_client() : CloudFrontClient {
...
+ * @return CloudWatchLogsClient
+ */
+function get_cloudwatch_logs_client() : CloudWatchLogsClient {
+	static $cloudwatch_logs_client;
+	if ( $cloudwatch_logs_client ) {
+		return $cloudwatch_logs_client;
+	}
+	$cloudwatch_logs_client = Altis\get_aws_sdk()->createCloudWatchLogs([
+		'version' => '2014-03-28',
+	]);
```
rmccue, line 875: Needs further spacing (not sure why that's not sniffed)

```suggestion
	$cloudwatch_logs_client = Altis\get_aws_sdk()->createCloudWatchLogs( [
		'version' => '2014-03-28',
	] );
```

Should we consider bumping this API version?

inc/namespace.php
```diff
@@ -924,9 +946,9 @@ function purge_cdn_paths( array $paths_patterns ) : bool {
...
 			'DistributionId'    => $distribution_id,
 			'InvalidationBatch' => [
-				'Paths'           => [
-					'Items'    => $paths_patterns,
-					'Quantity' => count( $paths_patterns ),
+				'Paths'         => [
+					'Items'     => $paths_patterns,
+					'Quantity'  => count( $paths_patterns ),
 				],
 				'CallerReference' => sha1( time() . wp_json_encode( $paths_patterns ) ),
```
rmccue, line 953: These actually no longer need to be aligned at all, and it's preferred that they aren't.

```suggestion
				'Paths' => [
					'Items' => $paths_patterns,
					'Quantity' => count( $paths_patterns ),
				],
				'CallerReference' => sha1( time() . wp_json_encode( $paths_patterns ) ),
```

inc/namespace.php
```diff
@@ -924,9 +946,9 @@ function purge_cdn_paths( array $paths_patterns ) : bool {
 		$client->createInvalidation( [
 			'DistributionId'    => $distribution_id,
 			'InvalidationBatch' => [
```
rmccue, line 948:
```suggestion
			'DistributionId' => $distribution_id,
			'InvalidationBatch' => [
```

inc/namespace.php
```diff
@@ -954,3 +976,95 @@ function set_wp_debug_constants() : void {
...
+
+/**
+ * Returns true when environment is running in the cloud. Returns false in all
+ * other conditions, such as local-server or local-chassis.
+ *
+ * @return bool
+ */
+function is_cloud() : bool {
+	return in_array( Altis\get_environment_architecture(), [ 'ec2', 'ecs' ], true );
+}
```
rmccue, line 988: This feels like maybe it should be in altis-core instead

inc/namespace.php
```diff
@@ -954,3 +976,95 @@ function set_wp_debug_constants() : void {
...
+	// correctly routed by our Fluent Bit container. Fluent Bit is configured
+	// to extract the log group and log stream from this string, with the log
+	// group in the second position and the log stream in the third,
+	// deliminated by a period. It will automatically prepend the environment
+	// name to the log group.
+	$tag_name = sprintf( 'app.%s.%s', $log_group, $log_stream );
+
+	// Let's store each logger in an array so that we don't keep instantiating
+	// loggers. We create a new logger for each Monolog channel. The channel
+	// name will be used as the Fluent Bit tag.
```
rmccue, line 1020: The commenting here is a bit too verbose IMO, and makes the code less readable. The first comment is great in terms of explaining _why_, but some of these just explain _what_, which you can already see from the code. Recommend slimming these down.

```suggestion
```

inc/namespace.php
```diff
@@ -954,3 +976,95 @@ function set_wp_debug_constants() : void {
...
+	static $loggers = [];
+	if ( isset( $loggers[ $tag_name ] ) ) {
+		return $loggers[ $tag_name ];
+	}
+
+	$logger = new Logger( $tag_name );
+
+	// Use Fluent Bit if it's available.
```
rmccue, line 1028:
```suggestion
```

inc/namespace.php
```diff
@@ -954,3 +976,95 @@ function set_wp_debug_constants() : void {
...
+	if ( Fluent_Bit\is_available() ) {
+		// Use Monolog's built-in TCP socket handler.
```
rmccue, line 1030:
```suggestion
```

inc/ses_to_cloudwatch/namespace.php
```diff
@@ -45,21 +35,11 @@ function on_sent_message( $result, $message ) {
...
-						'class'   => get_class( $error ),
-						'message' => $error->getMessage(),
-					],
-					'message' => $message,
-				] ),
-			],
+	Cloud\get_logger( 'ses', 'Failed' )->error( json_encode( [
+		'error'     => [
+			'class'   => get_class( $error ),
+			'message' => $error->getMessage(),
```
rmccue, line 41: As above, this shouldn't be aligned.
```suggestion
		'error' => [
			'class' => get_class( $error ),
			'message' => $error->getMessage(),
```

rmccue (changes requested):

rmccue:
Oop, forgot to add the review comment too.

The approach and code looks good to me, just some formatting and readability concerns.

inc/namespace.php
rmccue: (Not necessarily in this PR, as a follow up for the future.)

nathanielks:
@rmccue I've lost the comment asking if we should bump CloudWatch Logs' API version. It's already on the latest, so there's no need.

rmccue (approved):

PR #308, opened 2021-02-08
Remove reflect cloudfront headers function

roborourke:
This was dumb and I am dumb and this does literally nothing because you cannot read response headers client side without making the request from JS itself.


nathanielks (approved):
❤️

Issue #312, opened 2021-02-17
Getting URLs for S3 uploaded attachments involves repeatedly switching blogs 


goldenapples:
I've noticed that the "query-attachments" admin ajax request on subsites in a multisite network takes sometimes 10-20 seconds to return results. This is most noticeable when opening the media library on a non-primary site - especially when Smart Media is active, because the cropper will return attachment URLs for each defined crop size of all 40 attachments returned per page. 

The slowest part of this process by far is the `Altis\Cloud\set_s3_uploads_bucket_url_hostname` function, which is hooked on "upload_dir". This filter takes on average 28ms to run, and it runs for every crop size for every image in the query results. In the x-ray trace I'm looking at ([1-602d3f5d-6bda22f81ccdd1901f4b08d2](https://dashboard.altis-dxp.com/#/us-east-1/playstation-prod-01/xray/trace/1-602d3f5d-6bda22f81ccdd1901f4b08d2)) this is being called 320 times (40 images in the results x 8 crop sizes each). 

I suspect that the cache thrashing caused by repeatedly switching blogs as part of the [`get_main_site_url()` function](wp-includes/cache.php) may be causing this performance hit. We should be able to cache the main site URL when first retrieved rather than switching blogs each time we need to build an s3 bucket URL.

roborourke:
This is my fault, there was an expectation that tachyon & uploads URLs would just work and match the site's host name whereas previously they were configured from environment variables related to the stack, see issue #217 and PR #218 for context.

We definitely want to avoid switching blogs.

roborourke:
Thanks @goldenapples, this is now available in the following releases: 4.0.15, 5.0.7 and 6.0.2

goldenapples:
Thanks so much for helping out with this @roborourke!

PR #313, opened 2021-02-17
Cache S3 upload dirs

goldenapples:
Prevents cache thrashing by repeatedly switching to and back from the main blog when building Tachyon URLs for attachments in multisite subsites.

See #312 - work in progress, opening pull request to test only.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -567,51 +567,51 @@ function set_tachyon_hostname( string $tachyon_url ) : string {
 function set_s3_uploads_bucket_url_hostname( array $dirs ) : array {
 	$s3_uploads = S3_Uploads::get_instance();
 
-	$s3_upload_dirs = wp_cache_get( 's3_upload_dirs' );
+	$cached_upload_dirs = wp_cache_get( 's3_upload_dirs' );
 
-	if ( ! $s3_upload_dirs ) {
-		$primary_host = wp_parse_url( get_main_site_url(), PHP_URL_HOST );
-		$s3_host = wp_parse_url( $s3_uploads->get_s3_url(), PHP_URL_HOST );
-		$current_host = wp_parse_url( site_url(), PHP_URL_HOST );
-
-		if ( ! $primary_host ) {
-			trigger_error( sprintf( 'Error parsing main site URL: %s', esc_url_raw( get_main_site_url() ) ), E_USER_WARNING );
-			return $dirs;
-		}
-
-		if ( ! $s3_host ) {
-			trigger_error( sprintf( 'Error parsing S3 bucket URL: %s', esc_url_raw( $s3_uploads->get_s3_url() ) ), E_USER_WARNING );
-			return $dirs;
-		}
-
-		if ( ! $current_host ) {
-			trigger_error( sprintf( 'Error parsing site URL: %s', esc_url_raw( site_url() ) ), E_USER_WARNING );
-			return $dirs;
-		}
-
-		// To support 3rd party CDNs leave the host names as is if the direct
-		// amazonaws.com URL is in use.
-		if ( strpos( $s3_host, '.amazonaws.com' ) !== false ) {
-			return $dirs;
-		}
-
-		// Ensure uploads host at least matches primary site host.
-		if ( $s3_host !== $primary_host ) {
-			$dirs['url'] = str_replace( "://{$s3_host}", "://{$primary_host}", $dirs['url'] );
-			$dirs['baseurl'] = str_replace( "://{$s3_host}", "://{$primary_host}", $dirs['baseurl'] );
-		}
-
-		// Only do the replacement if the host name is not a subdomain of the S3 host.
-		if ( substr( $current_host, -1 * strlen( $primary_host ) ) !== $primary_host ) {
-			$dirs['url'] = str_replace( "://{$primary_host}", "://{$current_host}", $dirs['url'] );
-			$dirs['baseurl'] = str_replace( "://{$primary_host}", "://{$current_host}", $dirs['baseurl'] );
-		}
-
-		$s3_upload_dirs = $dirs;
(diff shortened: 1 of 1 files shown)


roborourke:
Updated the title to indicate Work In Progress - stops the checks from passing until you're ready for review.

inc/namespace.php
```diff
@@ -567,44 +567,51 @@ function set_tachyon_hostname( string $tachyon_url ) : string {
...
-	$primary_host = wp_parse_url( get_main_site_url(), PHP_URL_HOST );
-	$s3_host = wp_parse_url( $s3_uploads->get_s3_url(), PHP_URL_HOST );
-	$current_host = wp_parse_url( site_url(), PHP_URL_HOST );
+	$s3_upload_dirs = wp_cache_get( 's3_upload_dirs' );
 
-	if ( ! $primary_host ) {
-		trigger_error( sprintf( 'Error parsing main site URL: %s', esc_url_raw( get_main_site_url() ) ), E_USER_WARNING );
-		return $dirs;
-	}
+	if ( ! $s3_upload_dirs ) {
```
roborourke, line 572: You can reverse this logic and return early to avoid the large if block.

inc/namespace.php
```diff
@@ -567,44 +567,51 @@ function set_tachyon_hostname( string $tachyon_url ) : string {
...
+			$dirs['url'] = str_replace( "://{$primary_host}", "://{$current_host}", $dirs['url'] );
+			$dirs['baseurl'] = str_replace( "://{$primary_host}", "://{$current_host}", $dirs['baseurl'] );
+		}
 
-	// Only do the replacement if the host name is not a subdomain of the S3 host.
-	if ( substr( $current_host, -1 * strlen( $primary_host ) ) !== $primary_host ) {
-		$dirs['url'] = str_replace( "://{$primary_host}", "://{$current_host}", $dirs['url'] );
-		$dirs['baseurl'] = str_replace( "://{$primary_host}", "://{$current_host}", $dirs['baseurl'] );
+		$s3_upload_dirs = $dirs;
+		wp_cache_set( 's3_upload_dirs', $s3_upload_dirs );
```
roborourke, line 611: Pretty sure this will need to be keyed or grouped on at least the `$current_host` as it's dynamic for each site. Possibly on the S3 host too but pretty sure that's static.

roborourke (commented):
Will also be good to cache the value in the `get_main_site_url()` function too. Prevent other uses of it causing performance problems. The path can be appended manually once the site URL has been retrieved to avoid needing to use in the cache key. The logic for that is pretty simple.

https://github.com/humanmade/altis-cloud/pull/313/files#diff-a0f318343a79085d556c9f6050c1261041075f5129f00e908fce122d89e02febR480

inc/namespace.php
goldenapples: Yeah, that's one thing I wasn't sure about. I thought that the object cache would automatically key entries by the site ID, but I should really validate that assumption.

In any case here, this isn't even a persistant entry so there _shouldn't_ be any way for it to need further grouping?

inc/namespace.php
goldenapples: Good call, done in b7c9794.

inc/namespace.php
roborourke: As I understand it any call to the object cache will use redis which is persistent. Maybe I’m missing something though

goldenapples: :facepalm: Nope, that's my bad. For asome reason I had it in my mind that not specifying an expiration meant creating a non-persistant cache entry. In actuality its just the opposite - no expiration means eternal.

roborourke: Also fwiw I was wrong - caches are keyed for the current site unless a group is specified that's registered as a global cache group so using `wp_cache_set()` is fine.

goldenapples: I'm happy to stick with using a static variable though for now - there are a few possibilities I could think of where caching this value between requests might cause issues (the site URL could change, one site could serve requests for more than one hostname, http vs https, etc.), and in any case the problem I was trying to guard against was just that this happen multiple times in one request, not that it happens at all.

goldenapples:
I'm ready for review here. On the site I was testing this on, this change cuts the media library ajax request from 10s down to 1.6s, and avoids the Gateway timeouts I've been seeing on subsites on production.

inc/namespace.php
roborourke: Makes sense to me 👍  Lower risk this way too.

inc/namespace.php
```diff
@@ -477,7 +477,13 @@ function load_db() {
  * @return string
  */
 function get_main_site_url( string $path = '' ) : string {
-	return get_site_url( get_main_site_id( get_main_network_id() ), $path );
+	static $main_site_url = '';
```
roborourke, line 480: I think you can do this without assigning an initial value too, but the end result is the same. Just feels a little better to do a boolean check against `null` or something that returns a boolean.
```suggestion
	static $main_site_url;
```

inc/namespace.php
```diff
@@ -561,10 +567,19 @@ function set_tachyon_hostname( string $tachyon_url ) : string {
...
  * Ensure the S3 Uploads Bucket URL matches the current site hostname.
  *
+ * Because this filter involves switching blogs, the results are cached in a
+ * static variable for performance.
+ *
  * @param array $dirs Uploads directories array.
  * @return array
  */
 function set_s3_uploads_bucket_url_hostname( array $dirs ) : array {
+	static $cached_upload_dirs = null;
```
roborourke, line 577: The `= null` bit is redundant here as that's the value it'll have anyway before being assigned. I think we can make the implementation consistent between the 2 functions here and leave off the default values.
```suggestion
	static $cached_upload_dirs;
```

roborourke (approved):
Functionally this looks good but I have 2 minor suggestions just to make it consistent.

PR #326, opened 2021-02-21
Add travis config

roborourke:


roborourke:
Whoops, this works, though the existing test was made obsolete with the fluent bit update. Needs removing.

roborourke:
There are no tests (that are in date) in this repo right now so there's no real need to add this but I'll use it as documentation for other modules. We can remove the obsolete test in a future update.

PR #334, opened 2021-03-05
Only use APCU if loaded

rmccue:
Our vantage-backend servers don't have APCU loaded, so this causes X-Ray to crash currently.


roborourke (approved):
Whoops.

roborourke:
Noting here for future ref, if something is going to be backported, it needs backporting to all supported versions as part of the process. I'll do it this time.

PR #362, opened 2021-04-16
Fix redis healthcheck

roborourke:
This is an effort to make the healthcheck backwards compatible in case wp-redis changes the stats method behaviour.

### DIFF
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -60,7 +60,7 @@ function output_page( array $checks ) {
 	if ( ! empty( $_SERVER['HTTP_ACCEPT'] ) && $_SERVER['HTTP_ACCEPT'] === 'application/json' ) {
 		$format = 'json';
 	}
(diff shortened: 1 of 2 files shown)


inc/healthcheck/namespace.php
```diff
@@ -60,7 +60,7 @@ function output_page( array $checks ) {
 	if ( ! empty( $_SERVER['HTTP_ACCEPT'] ) && $_SERVER['HTTP_ACCEPT'] === 'application/json' ) {
 		$format = 'json';
 	}
-	if ( $_GET['_accept'] ?? '' === 'json' ) {
+	if ( isset( $_GET['_accept'] ) && $_GET['_accept'] === 'json' ) {
```
rmccue, line 63: Why is this changed?

roborourke: Was getting some linter errors but I can change it back

jazzsequence (approved):

Issue #371, opened 2021-04-20
SES to Cloudwatch fails when email has attachment bigger than 64Kb


ivankristianto:
So while debugging the issues on intermittent errors with email not delivered.
See this ticket:
- https://dashboard.altis-dxp.com/#/support/ticket/4873/

I suspect this is because monolog use UDP when sending the package, max packet size is 65,535 bytes.

Steps to reproduce:
1. Go to any environments in Altis cloud, from the CLI command, try send wp_mail with attachment larger than 64Kb.
2. No email has been logged in the Email log altis dashboard

Expected:
I expect to see email is logged, success/failed, without attachment if it is a limitation.

Screenshots:
I have bunch of screenshots in this ticket: https://dashboard.altis-dxp.com/#/support/ticket/4873/

Acceptance criteria:

- [ ] ...

rmccue:
Updated to remove client details.

PR #374, opened 2021-04-21
Truncate SES message data to CloudWatch

ivankristianto:
Fixes: #371 

This PR is similar with the PR I made in https://github.com/humanmade/aws-xray/pull/77
With solve the same issue where the data send to cloudwatch could be more than 64Kb and thus we cannot see the email log if the email has attachment.

For logging purpose we don't need to store email attachment in the cloudwatch, it's enough information to see if the email header and delivery status.

So this PR truncate the message down.
Let me know what you think? @rmccue @roborourke @joehoyle


inc/ses_to_cloudwatch/namespace.php
```diff
@@ -43,3 +53,21 @@ function on_error_sending_message( Exception $error, $message ) {
...
+ * Truncate string to given maximum size.
+ *
+ * @param string $message     String message to truncate.
+ * @param int    $max_size    Maximum size in bytes, default 5KB.
+ * @param string $replacement String replacement.
+ *
+ * @return string
+ */
+function truncate_message( string $message, int $max_size = 5 * 1024, string $replacement = '…' ) : string {
+	if ( mb_strlen( $message ) < $max_size ) {
```
joehoyle, line 67: If you are counting length to be a proxy for size, this should be `strlen` not `mb_strlen`. A string that is `mb_strlen` of 1 can be a size of 3bytes for example.

joehoyle (changes requested):

inc/ses_to_cloudwatch/namespace.php
```diff
@@ -43,3 +53,21 @@ function on_error_sending_message( Exception $error, $message ) {
...
+/**
+ * Truncate string to given maximum size.
+ *
+ * @param string $message     String message to truncate.
+ * @param int    $max_size    Maximum size in bytes, default 5KB.
+ * @param string $replacement String replacement.
+ *
+ * @return string
+ */
+function truncate_message( string $message, int $max_size = 5 * 1024, string $replacement = '…' ) : string {
```
joehoyle, line 66: This seems maybe `truncate_string` is more appropriate, as from the perspective of this function, it doesn't need to be a "message"

joehoyle (approved):

joehoyle:
Handing off to @roborourke for final review

inc/ses_to_cloudwatch/namespace.php
```diff
@@ -43,3 +53,21 @@ function on_error_sending_message( Exception $error, $message ) {
 		'message' => $message,
 	] ) );
 }
+
+/**
+ * Truncate string to given maximum size.
+ *
+ * @param string $string      String message to truncate.
+ * @param int    $max_size    Maximum size in bytes, default 5KB.
+ * @param string $replacement String replacement.
```
roborourke, line 62: We don't align docblock params any more, just prevents unnecessary code changes when adding / changing a parameter in future.
```suggestion
 * @param string $string String to truncate.
 * @param int $max_size Maximum size in bytes, default 5KB.
 * @param string $replacement String replacement.
```

inc/ses_to_cloudwatch/namespace.php
```diff
@@ -43,3 +53,21 @@ function on_error_sending_message( Exception $error, $message ) {
...
+ *
+ * @return string
+ */
+function truncate_string( string $string, int $max_size = 5 * 1024, string $replacement = '…' ) : string {
+	if ( strlen( $string ) < $max_size ) {
+		return $string;
+	}
+
+	// Truncate string in the middle.
+	return substr_replace( $string, $replacement, $max_size / 2, strlen( $string ) - $max_size + strlen( $replacement ) );
```
roborourke, line 72: Took me a while to grok what this is doing! Nice work

roborourke (approved):

PR #384, opened 2021-04-23
Fix linter errors

jazzsequence:
Fixes linting errors in `altis/cloud`

### DIFF
--- a/assets/environment-indicator.css
+++ b/assets/environment-indicator.css
@@ -7,7 +7,7 @@
 }
 
 #wpadminbar .altis-env-indicator {
-	background-color: #4667DE;
+	background-color: #4667de;
 	display: inline-block;
 	height: 1.2em;
 	margin-top: 6px;
--- a/assets/environment-indicator.css
+++ b/assets/environment-indicator.css
@@ -23,6 +23,6 @@
 }
 
 #wpadminbar #wp-admin-bar-altis .dashicons-before::before {
-	font-size: 1.2em;
+	font-size: 16px; /* stylelint-disable-line declaration-property-unit-blacklist */
 	vertical-align: middle;
 }
--- a/assets/environment-indicator.css
+++ b/assets/environment-indicator.css
@@ -23,6 +23,6 @@
 }
 
 #wpadminbar #wp-admin-bar-altis .dashicons-before::before {
-	font-size: 16px;
+	font-size: 1.2em;
 	vertical-align: middle;
 }
--- a/assets/environment-indicator.css
+++ b/assets/environment-indicator.css
@@ -10,15 +10,12 @@
 	background-color: #4667de;
 	display: inline-block;
 	height: 1.2em;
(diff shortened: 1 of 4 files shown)


inc/page_cache/namespace.php
```diff
@@ -80,7 +80,7 @@ function get_unique_headers() : array {
 		$header_key = 'HTTP_' . str_replace( '-', '_', strtoupper( $header ) );
 		if ( isset( $_SERVER[ $header_key ] ) ) {
 			// Add header to batcache vary keys.
-			$unique_keys[ $header ] = $_SERVER[ $header_key ];
+			$unique_keys[ $header ] = wp_unslash( $_SERVER[ $header_key ] ); // phpcs:ignore HM.Security.ValidatedSanitizedInput.InputNotSanitized
```
roborourke, line 83: I'm pretty sure `wp_unslash()` won't be available yet as this runs super early during the bootstrap process. For long lines like this you can also put the `phpcs:ignore` rule on the line above.

inc/page_cache/namespace.php
```diff
@@ -100,7 +100,7 @@ function get_unique_cookies() : array {
 	foreach ( $unique_cookies as $cookie ) {
 		if ( ! empty( $_COOKIE[ $cookie ] ) ) {
 			// Add header to batcache vary keys.
-			$unique_keys[ $cookie ] = $_COOKIE[ $cookie ];
+			$unique_keys[ $cookie ] = wp_unslash( $_COOKIE[ $cookie ] ); // phpcs:ignore HM.Security.ValidatedSanitizedInput.InputNotSanitized
```
roborourke, line 103: Same here.

assets/environment-indicator.css
```diff
@@ -7,25 +7,22 @@
...
-	line-height: 1.2;
-}
-
 #wpadminbar .altis-env-indicator abbr {
 	line-height: inherit;
 }
 
 #wpadminbar #wp-admin-bar-altis .dashicons-before::before {
-	font-size: 16px;
+	font-size: 1.2em;
```
roborourke, line 26: Our stylelint rules don't necessarily match the way the WP admin or the admin bar is styled so I'd make absolutely certain this is going to work consistently. I've generally opted for using stylelint ignore rules rather than changing the units.

roborourke (commented):
I think the `wp_unslash()` references added definitely need changing and wanted to double check the behaviour of the font-size unit change. While we enforce those style rules for our own original code it's important to stay consistent with how the WP admin is styled which tends to use pixel font sizes everywhere.

assets/environment-indicator.css
jazzsequence: I tested this locally to make sure I got the sizing correct. This works, but if it's better to match the existing (bad) rules, I can ignore it in the stylelint rules.

roborourke (approved):

PR #397, opened 2021-05-29
Link directly to the X-Ray trace

rmccue:
This allows viewing the request trace directly, useful if you're on production without the dev tools enabled.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -27,6 +27,7 @@ use Monolog\Logger;
 use Psr\Http\Message\RequestInterface;
 use Psr\Log\LoggerInterface;
 use S3_Uploads;
+use WP_Admin_Bar;
 use WP_CLI;
 
 /**
@@ -64,6 +65,7 @@ function bootstrap() {
 	) {
 		require_once Altis\ROOT_DIR . '/vendor/humanmade/aws-xray/inc/namespace.php';
 		require_once Altis\ROOT_DIR . '/vendor/humanmade/aws-xray/plugin.php';
+		add_action( 'admin_bar_menu', __NAMESPACE__ . '\\register_trace_menu_item', 100 );
 		add_filter( 'aws_xray.redact_metadata', __NAMESPACE__ . '\\remove_xray_metadata' );
 		if ( in_array( Altis\get_environment_architecture(), [ 'ec2', 'ecs' ], true ) ) {
 			add_filter( 'aws_xray.trace_to_daemon', __NAMESPACE__ . '\\add_ec2_instance_data_to_xray' );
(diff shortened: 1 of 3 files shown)


inc/namespace.php
```diff
@@ -1084,3 +1086,32 @@ function get_logger( string $log_group, string $log_stream ) : LoggerInterface {
...
+function register_trace_menu_item( WP_Admin_Bar $wp_admin_bar ) {
+	if ( ! defined( 'HM_ENV_REGION' ) ) {
+		return;
+	}
+
+	$trace_id = XRay\get_root_trace_id();
+	$url = sprintf(
+		'https://dashboard.altis-dxp.com/#/%s/%s/xray/trace/%s',
+		HM_ENV_REGION,
+		Altis\get_environment_name(),
```
roborourke, line 1104: This can return "unknown" locally.

inc/namespace.php
```diff
@@ -1084,3 +1086,32 @@ function get_logger( string $log_group, string $log_stream ) : LoggerInterface {
...
 	return $loggers[ $tag_name ];
 }
+
+/**
+ * Register the "Debug Request" menu item.
+ *
+ * @param WP_Admin_Bar $wp_admin_bar Admin bar instance.
+ */
+function register_trace_menu_item( WP_Admin_Bar $wp_admin_bar ) {
+	if ( ! defined( 'HM_ENV_REGION' ) ) {
```
roborourke, line 1096: This isn't really sufficient to check for a cloud environment, there's an `is_cloud()` function in this namespace.

roborourke (changes requested):
Couple of questions:

- What's the reason for putting this under the Altis admin menu?
- Who needs to see this?

rmccue:
> What's the reason for putting this under the Altis admin menu?

Gotta go somewhere; we already have meta-links to the Altis Dashboard from here, so it felt the most appropriate place to put it. It also ensures that it's not using up visual space for most users who don't need it.

> Who needs to see this?

I considered tying it to super admin or something, but I think I'd rather go broad here as there's no real reason to limit it apart from visual noise/irrelevancy. Given it's under the Altis menu where we have those other links anyway, that's not a big issue.

inc/namespace.php
rmccue: I did wonder, but this is the same check as in the environment indicator code:

https://github.com/humanmade/altis-cloud/blob/deac1cc0d8aeeda006d8ccc38f136330596bf5d8/inc/environment_indicator/namespace.php#L93-L100

roborourke: Hmm. That's a little different to be fair, it can still be useful to jump to the dashboard from your local, but not to jump to a specific xray entry in the dashboard from your local.

roborourke:
Ok, might be useful to document it then at least.

Main reason I asked about the location is because when we do have Query Monitor available are we hiding what is arguably a useful link somewhere unexpected when it could be part of that menu?

Would it be useful to add these links as response headers for REST / Ajax requests too?

rmccue:
> Ok, might be useful to document it then at least.

👍 - I actually thought of this as I was thinking through how to create a video about debugging.

> Main reason I asked about the location is because when we do have Query Monitor available are we hiding what is arguably a useful link somewhere unexpected when it could be part of that menu?

Good Q; we could add it to both. Can link across from the screen which currently just has a rando dump of X-Ray data.

> Would it be useful to add these links as response headers for REST / Ajax requests too?

_Possibly_; just need to check auth I think. Can add that too. It's a bit less discoverable, but can document it.

roborourke (approved):

roborourke:
Ignored the AJAX request idea for now. I think QM has a good way of doing it but will add in a follow up.

PR #428, opened 2021-07-30
Fix early uploads bad usage

### LINKED ISSUE #421: Warn when using `wp_upload_dir()` too early
S3 Uploads is loaded on the `plugins_loaded` hook early so we need to warn developers trying to use `wp_upload_dir()` before S3 uploads has rewritten the paths.

Acceptance criteria:

- [x] Warn using `_doing_it_wrong()` if S3 uploads has not been initialised when calling `wp_upload_dir()`
- [x] Avoid caching uploads directory in `Altis\Cloud\set_s3_uploads_bucket_url_hostname` if called too early
- [x] Avoid caching uploads directory in `Altis\Cloud\set_s3_uploads_bucket_url_hostname` if the directory is a local path instead of an S3 path

rmccue:
Fixes #421.

Also now activated on all environments to ensure consistency. This will cause the uploads URL on Local Server to change from (eg) `https://s3-altis-dev.altis.dev/uploads` to `https://altis-dev.altis.dev/uploads`.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -592,6 +592,13 @@ function set_tachyon_hostname( string $tachyon_url ) : string {
  * @return array
  */
 function set_s3_uploads_bucket_url_hostname( array $dirs ) : array {
+	// Has S3 Uploads been initialized yet?
+	$wrappers = stream_get_wrappers();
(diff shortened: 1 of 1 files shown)


roborourke (approved):
Guess this needs backporting too

PR #440, opened 2021-08-06
Provide PHP session handler via Redis

shadyvb:
humanmade/altis-core#96

- Add PHP Sessions support via Redis
- Register dummy session handler by default if not enabled
- Add doc to activate the new feature via config


inc/namespace.php
```diff
@@ -167,6 +169,15 @@ function load_platform( $wp_debug_enabled ) {
 		load_object_cache_redis();
 	}
 
+	// Session support.
+	require_once __DIR__ . '/session_handler/namespace.php';
```
roborourke, line 173: For files that don't have side effects we add those to the module's `autoload.files` section in the `composer.json` rather than requiring directly. When testing locally you might need to commit the change before running `composer update` to get the updated autoload files.

inc/namespace.php
```diff
@@ -427,6 +438,36 @@ function load_object_cache_redis() {
...
+	if ( headers_sent() ) {
+		return;
+	}
+
+	$client = new Predis\Client( $wp_object_cache->build_client_parameters( $redis_server ), [ 'prefix' => 'sessions:' ] );
+	$handler = new Predis\Session\Handler( $client );
+
+	session_set_save_handler( $handler, true );
+	session_name( 'altis_session' );
+	session_start();
```
roborourke, line 458: Looks like we don't need to call `session_start()` ourselves based on the slack conversation, it's up to developers to do so. Might be worth extending the documentation page a bit to indicate that and what the default cookie name is.

inc/session_handler/class-disallowed-session-handler.php
```diff
@@ -0,0 +1,29 @@
...
+// phpcs:disable Squiz.Commenting.FunctionComment.Missing
+
+/**
+ * Disallowed Session handler class.
+ *
+ * Disables creation of sessions via short-circuiting session_open() calls.
+ */
+class Disallowed_Session_Handler implements \SessionHandlerInterface {
+	function open( $save_path, $session_name ) {
+		error_log( 'Sessions are not enabled in Altis config.' );
```
roborourke, line 21: Use `trigger_error( '...', E_USER_WARNING );` instead. `error_log()` just writes to log file but doesn't raise any kind of warning or error and doesn't have a severity level. Using `trigger_error()` means tools like Query Monitor will pick up and show the warnings.

inc/session_handler/namespace.php
```diff
@@ -0,0 +1,8 @@
+<?php
+/**
+ * Altis Cloud Session handler.
+ *
+ * @package altis/cloud
+ */
+
+require_once __DIR__ . '/class-disallowed-session-handler.php';
```
roborourke, line 8: I think this maybe counts as a side-effect 🤔  Either way this require can be moved to the module's `autoload.classmap` property in the `composer.json`.

Follow the pattern set out here:

https://github.com/humanmade/altis-cloud/blob/master/composer.json#L12-L27

roborourke (changes requested):
Looking solid, just a couple of things to update.

inc/namespace.php
shadyvb: I was wondering why I didn't see any require calls for other files 🤦

composer.json
```diff
@@ -51,4 +52,4 @@
 			]
 		}
 	}
-}
+}
```
roborourke, line 55: I have my editor set to add a newline at the end of files, not a blocker but I think it's something we default to more broadly in the coding standards, it avoids superfluous changes like this one.

composer.json
```diff
@@ -20,6 +20,7 @@
 			"inc/environment_indicator/namespace.php",
 			"inc/cloudfront_media_purge/namespace.php",
 			"inc/fluent_bit/namespace.php",
+			"inc/session_handler/class-disallowed-session-handler.php",
```
roborourke, line 23: I never noticed that we had `inc/class-loggerexceptionhandler.php` below. Strictly speaking that should be in the `classmap` section above. It could be converted to the following and the class files could be removed from this files section:

```
"classmap": [
	"inc/"
],
```

We can deal with that in a follow up I think as it might make backporting difficult.

roborourke (approved):

docs/sessions.md
```diff
@@ -0,0 +1,19 @@
...
+				"cloud": {
+					"redis": true
+				}
+			}
+		}
+	}
+}
+```
+
+Worth noting is that the cookie name is `altis_session`, rather than the default `PHPSESSID`.
```
roborourke, line 19: There's a general pattern in the docs like `**Note:** ...` that we could use here, for consistency.

```suggestion
**Note:** The cookie name is `altis_session`, rather than the default `PHPSESSID` in order to bypass CloudFront page caching.
```

composer.json
shadyvb:
```suggestion
}

```

roborourke:
Sorry one last question here, do we know what happens if the `php.ini` setting is still a blank string? Given that infrastructure updates aren't tied to the software version directly I want to be sure things won't break in that event.

shadyvb:
@roborourke In that situation, sessions will be disabled, and won't turn on to begin with, so won't break anything.

PR #445, opened 2021-08-18
Allow instance healthcheck to run before WP or plugins load

### LINKED ISSUE #438: Run instance healthcheck before WordPress loads
Our instance healthcheck (`/__instance_healthcheck/`) is intended to check that PHP is loading correctly, but not that WP can load up fully. This needs to trigger extremely early in the loading process (likely as soon as modules load); the regular healthcheck should be later in the process.

Importantly, this needs to work across all hostnames, including our internal load balancer names. Currently, it is generating a 302 as WP can't route the host name.

Acceptance criteria:
* [x] Accessing the instance healthcheck with an arbitrary hostname should return a 200 with the healthcheck data (not a 302)
* [x] If a plugin contains a fatal error, the instance healthcheck should still return a 200 with the healthcheck data

roborourke:
Fixes #438

### DIFF
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -124,7 +124,7 @@ function output_page( array $checks ) {
 			<?php foreach ( $checks as $check => $status ) : ?>
 				<tr>
 					<td>
-						<?php echo htmlspecialchars( $check ) ?>
+						<?php echo htmlspecialchars( $check ) // phpcs:ignore HM.Security.EscapeOutput.OutputNotEscaped ?>
 					</td>
 					<td>
 						<?php echo is_error( $status ) ? sprintf( '%s (code: %s)', esc_html( $status->get_error_message() ), esc_html( $status->get_error_code() ) ) : 'OK' ?>
--- a/inc/healthcheck/namespace.php
+++ b/inc/healthcheck/namespace.php
@@ -26,18 +26,51 @@ function bootstrap() {
 		return;
 	}
 
-	if ( strpos( $_SERVER['REQUEST_URI'], '/__healthcheck' ) !== false ) {
-		output_page( run_checks() );
-		return;
-	}
+	// Defer app level healthchecks until after plugins have laoded.
+	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_healthcheck', 100 );
 
+	// Run instance healthcheck immediately.
 	if ( strpos( $_SERVER['REQUEST_URI'], '/__instance_healthcheck' ) !== false ) {
 		output_page( run_instance_checks() );
 	}
 }
 
 /**
+ * Loads the application healthcheck.
+ *
+ * @return void
+ */
+function load_healthcheck() {
+	if ( defined( 'WP_INSTALLING' ) && WP_INSTALLING ) {
+		return;
+	}
+
+	Cavalcade\bootstrap();
+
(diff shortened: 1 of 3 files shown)


inc/healthcheck/namespace.php
```diff
@@ -26,35 +26,75 @@ function bootstrap() {
...
+ *
+ * @param mixed $value The variable to check.
+ * @return boolean
+ */
+function is_error( $value ) : bool {
+	if ( function_exists( 'is_wp_error' ) ) {
+		return is_wp_error( $value );
 	}
+	return $value !== true;
 }
```
shadyvb, line 67: `is_wp_error` is simple enough to replicate, maybe we should go for the instanceof check here ?

```suggestion
function is_error( $value ) : bool {
	return $value instanceof WP_Error;
}
```

shadyvb (approved):

rmccue (approved):

inc/healthcheck/namespace.php
rmccue: I was thinking the same, but presumably the class isn't loaded?

shadyvb: I think bunch of the checks are already assuming it exists, eg: `run_mysql_healthcheck`. @roborourke can you clarify the use case for this change ?

roborourke: I went for this approach in case we needed to add additional checks to the instance healthcheck in future. If we do those can't return a `WP_Error` but might need some sort of failed status still. Given our other healthchecks return `true` or a `WP_Error` I felt this was the most straight forward to account for alternative healthcheck return values before WP has loaded.

roborourke: Also not sure if the `WP_Error` class would have to exist to be able to do an instanceof check against it 🤔

shadyvb: Fair! I take back my suggestion then.

roborourke: TIL that can be done, the class doesn't need to exist. Still though, needs to fall back to checking if the value is `true` or not otherwise an early health check could return false and still count as a pass.

inc/healthcheck/namespace.php
```diff
@@ -26,35 +26,75 @@ function bootstrap() {
 		return;
 	}
 
+	// Defer app level healthchecks until after plugins have laoded.
+	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_healthcheck', 100 );
+
+	// Run instance healthcheck immediately.
+	if ( strpos( $_SERVER['REQUEST_URI'], '/__instance_healthcheck' ) !== false ) {
```
rmccue, line 33: Oh, maybe this should have been `=== 0` btw @roborourke?

rmccue: I guess was just same as before

roborourke: Mayyybe. Could it ever be the case that Altis is served from a subdirectory? I know we don't really support that. Seems like something that's not really gonna present a problem given the double underscore prefix.

rmccue: > Could it ever be the case that Altis is served from a subdirectory?

No, it'll crash the server iirc if you do (including with the existing healthcheck).

> Seems like something that's not really gonna present a problem given the double underscore prefix.

Yeah, just thinking of what the "best" is

PR #455, opened 2021-09-15
Check inline index settings option to bypass package creation

roborourke:
Related to https://github.com/humanmade/product-dev/issues/767

### DIFF
--- a/inc/elasticsearch_packages/namespace.php
+++ b/inc/elasticsearch_packages/namespace.php
@@ -27,12 +27,25 @@ function bootstrap() : void {
 		return;
 	}
 
+	$should_use_inline_index_settings = Altis\get_config()['search']['inline-index-settings'] ?? false;
+	$use_inline_index_settings = (
+		$should_use_inline_index_settings &&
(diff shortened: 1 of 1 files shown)


shadyvb:
Hmmmm, trying to think what happens to existing data on activation of inline settings 🤔 Do we need to consider a migration path here ?

roborourke:
Good shout. I guess the changes would only kick in when a site is re-indexed, it doesn't modify the queries being run at all so that part of it shouldn't be affected.

We could note that a reindex is recommended after upgrading the ES version in the upgrade guide.

roborourke:
Ohh, I see yeah. Given that it's a config option that could be problematic if the ES packages don't exist yet, we have no command or way to create them after the fact.

I guess the hybrid model is our best bet for keeping things compatible. With that being the case we could even programmatically handle the switch between inline and stored file 🤔 

I'll close out this PR anyway. I think I was originally envisaging this as not using the packages API at all with ES 7 but your approach is better.

shadyvb:
In hindsight, I should've explained my question better 😅 that it is about the package data not the indexed data. But yeah, glad we aligned eventually.

PR #467, opened 2021-09-30
Update the Composer version number

BronsonQuick:
The Composer version number is incorrect.
Composer version output


roborourke:
I think we may need to check the build container that's running, we haven't officially rolled out Composer v2 yet so the recommendation is to add `composer self-update --2` to your build script.

We will be upgrading the default composer version soon, possibly alongside Altis v9.

roborourke:
Not a pain at all! It was timely really as we need to update the build containers default versions of node etc... soon anyway. We've got an epic in the backlog so hopefully we'll get things moving there soon. No stress mate :)

PR #487, opened 2021-11-15
Rework page caching documentation

rmccue:


docs/page-caching.md
```diff
@@ -1,42 +1,74 @@
...
+- Server errors (`5XX`): 300 seconds (5 minutes)
+
+Time-to-live values should be balanced between achieving high cache hit rates and ensuring fresh content is served.
+
+This can be adjusted by sending a [`Cache-Control: max-age=x` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives) within your response, which will override any defaults.
+
+For example, to specify a lifetime of one hour for a specific page:
+
+```php
+header( 'Cache-Control: max-age=' . HOUR_IN_SECONDS );
```
joehoyle, line 54: This header is set by batcache, and I think we should recommend changing the batcache config rather than sending a header

joehoyle (changes requested):

docs/page-caching.md
rmccue: Batcache [parses it from the header](https://github.com/humanmade/batcache/blob/3230d4424556b3a1c4209d7548a2f2d9a2927907/advanced-cache.php#L275)

joehoyle: Yeah I realise that, but it kinda sucks!

rmccue: Eh, I think it works just fine. If you manually set `$batcache->max_age`, it'll get clobbered by that anyway. Also, this is better than using the weird undocumented `$batcache` global half way through a page render.

joehoyle: Yeah, but like the comment says, you have to also check when you can send headers, so it's pretty rough. I think the best solution is `page-cache.ttl` in the config. Baring that, I'd still go `$batcache` global over having to send output just to tell the cache to do something.

rmccue: I was documenting this specifically for a use case which requires dynamic setting of the max age based on the content (i.e. cache any posts older than a month for a full day instead of 5 mins).

You actually _don't_ have to check when you can send headers since the output is all buffered anyway, but it's best practice in PHP. If that's a sticking point, I can just remove it.

joehoyle: Ok, imo it's a worse api, but if we go with that I don't have an issue with the dev warning.

joehoyle (approved):

docs/page-caching.md
```diff
@@ -93,39 +137,51 @@ global $batcache;
...
+## Cache Invalidation
 
-## Debugging
+Caches are automatically invalidated when their time-to-live expires.
 
-Its not uncommon to need debug why things may or may not be served from the page cache. There's 2 response headers that indicate the cache status, and reasons:
+The CDN cache can be [manually purged](https://docs.altis-dxp.com/cloud/cdn/purge/) through a function call. This should be used sparingly, as it will negatively affect cache hit rates, and incurs additional stress on the CDN system. This functionality is monitored, and excessive use may be considered a violation of your terms of service.
 
-### X-Batcache Header
+It is not currently possible to clear Batcache cache values manually. As Batcache cache values are stored within the object cache, the object cache can be cleared if Batcache needs to be cleared, however this also must be used sparingly.
```
rmccue, line 147: We should probably add this ability, btw.

owaincuvelier (approved):

owaincuvelier:
LGTM!

PR #488, opened 2021-11-15
:white_check_mark: Check support ticket link works properly

shadyvb:
Requires https://github.com/humanmade/altis-dev-tools/pull/231

Fixes https://github.com/humanmade/product-dev/issues/839

### DIFF
--- a/tests/acceptance/AdminCest.php
+++ b/tests/acceptance/AdminCest.php
@@ -37,7 +37,7 @@ class AdminCest {
 	 */
 	public function testSupportTicketLinkWithCloud( AcceptanceTester $I ) {
 		// Schedule running of _configSupportTicketLinkWithCloud in next Acceptance Test.
-		$rollback = $I->bootstrapWith( __FILE__, [ __CLASS__, '_configSupportTicketLinkWithCloud' ] );
+		$rollback = $I->bootstrapWith( [ __CLASS__, '_configSupportTicketLinkWithCloud' ] );
 
 		$I->wantToTest( 'Support ticket link is correct.' );
 		$I->loginAsAdmin();
--- a/tests/acceptance/AdminCest.php
+++ b/tests/acceptance/AdminCest.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Tests for core module's admin features.
+ *
+ * phpcs:disable WordPress.Files, WordPress.NamingConventions, PSR1.Classes.ClassDeclaration.MissingNamespace, HM.Functions.NamespacedFunctions
+ */
+
+use Codeception\Util\Locator;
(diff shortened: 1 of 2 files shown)


tests/acceptance/AdminCest.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Test module versions are displayed correctly on about page.
+	 *
+	 * @param AcceptanceTester $I Tester
+	 *
+	 * @return void
+	 */
+	public function testSupportTicketLinkWithCloud( AcceptanceTester $I ) {
+		// Schedule running of _configSupportTicketLinkWithCloud in next Acceptance Test.
+		$rollback = $I->preLoadConfig( __FILE__, [ __CLASS__, '_configSupportTicketLinkWithCloud' ] );
```
roborourke, line 40: Do you know if each test method is a new session? Would it be possible sense to use the `_before()` and `_after()` methods for example? I guess that would affect every test in the file.

roborourke (approved):

tests/acceptance/AdminCest.php
shadyvb: I'm returning the function to be called by the method at the end, so it clears it after the work is done. But in case the method doesn't do that, it'll be cleared on shutdown.

We could in theory do a before and after, and use some naming convention similar to how data sources work in phpunit, but it seemed a bit excessive to do so. I guess we'll know how much we'll need that as write more tests.

FYI that we'll need something very similar to update composer config, unless we have a filter that could be used with `preloadConfig`, eg: to activate authorship, but I think that might be a bit late, will see.

roborourke: Hm yeah, was wondering about that. The config can be filtered and will start being used after the `altis.modules.init` action hook so I think this same approach could work.

This looks good, I'm just thinking through some ideas on the other PR that adds the `preLoadConfig` method.

tests/acceptance/AdminCest.php
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * @param AcceptanceTester $I Tester
+	 *
+	 * @return void
+	 */
+	public function testSupportTicketLinkWithCloud( AcceptanceTester $I ) {
+		// Schedule running of _configSupportTicketLinkWithCloud in next Acceptance Test.
+		$rollback = $I->bootstrapWith( [ __CLASS__, '_configSupportTicketLinkWithCloud' ] );
+
+		$I->wantToTest( 'Support ticket link is correct.' );
```
roborourke, line 42:
```suggestion
		$I->wantToTest( 'Support ticket link with application is correct.' );
```

PR #511, opened 2021-11-23
Use main network domain as from address

### LINKED ISSUE #502: Default From address for emails is set to an uncontrolled domain
Steps to reproduce:
1. Create a new Altis site
2. Attempt to send an email (eg reset password)
3. Observe error with sending due to unverified domain

The From address should default to `no-reply@{main domain}` as with WordPress core, rather than `no-reply@humanmade.com`

Acceptance criteria:

- [x] From address defaults to `no-reply@{main domain}` where `{main domain}` is the network domain
- [x] From address can be overridden within the composer.json configuration

roborourke:
Fixes #502

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -81,12 +81,16 @@ function bootstrap() {
 		define( 'WP_CACHE', true );
 	}
 
-	add_filter( 'wp_mail_from', function ( string $email ) use ( $config ) : string {
+	add_filter( 'wp_mail_from', function () use ( $config ) : string {
+		$default_email = sprintf(
+			'noreply@%s',
+			wp_parse_url( network_home_url( '/' ), PHP_URL_HOST )
+		);
 		return filter_var(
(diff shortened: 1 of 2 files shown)


shadyvb (approved):

inc/namespace.php
```diff
@@ -81,12 +81,16 @@ function bootstrap() {
 		define( 'WP_CACHE', true );
 	}
 
-	add_filter( 'wp_mail_from', function ( string $email ) use ( $config ) : string {
+	add_filter( 'wp_mail_from', function () use ( $config ) : string {
+		$default_email = sprintf(
+			'noreply@%s',
+			wp_parse_url( network_home_url( '/' ), PHP_URL_HOST )
```
rmccue, line 87: This should probably exclude `www` by the way, not sure if it does by default?

roborourke: Ah crap. Will do a follow up.

PR #516, opened 2021-11-24
Strip www. from default from mail address

roborourke:
Follow on to #511

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -82,10 +82,11 @@ function bootstrap() {
 	}
 
 	add_filter( 'wp_mail_from', function () use ( $config ) : string {
-		$default_email = sprintf(
-			'noreply@%s',
-			str_replace( 'www.', '', wp_parse_url( network_home_url( '/' ), PHP_URL_HOST ) )
-		);
+		$domain = get_network()->domain;
+		if ( substr( $domain, 0, 4 ) === 'www.' ) {
+			$domain = substr( $domain, 4 );
+		}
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -84,7 +84,7 @@ function bootstrap() {
 	add_filter( 'wp_mail_from', function () use ( $config ) : string {
 		$default_email = sprintf(
 			'noreply@%s',
-			wp_parse_url( network_home_url( '/' ), PHP_URL_HOST )
+			str_replace( 'www.', '', wp_parse_url( network_home_url( '/' ), PHP_URL_HOST ) )
```
rmccue, line 87: This will replace it anywhere in the URL, so if we had a client like `bestofwww.com` it'd ruin the domain.

I think actually though, using something like this may be better:
```php
$domain = get_network()->domain;
if ( substr( $domain, 0, 4 ) === 'www. ) {
    $domain = substr( $domain, 4 );
}
```

No need to generate a full URL.

joehoyle:
This seems like it would be a backward compat issue? If someone has www.mydomain.com verified, doing this update would change the default and break.

roborourke:
@joehoyle can we look at all existing settings at a glance and determine if it's a breaking change or not?

rmccue:
> This seems like it would be a backward compat issue? If someone has [www.mydomain.com](http://www.mydomain.com) verified, doing this update would change the default and break.

The default up until yesterday was `noreply@humanmade.com`, we only just changed it in #511, so should be no major diff.

rmccue (approved):

PR #521, opened 2021-11-24
Check if this is intial install before using `get_network()`

roborourke:
Multisite functions are not available during the initial install.


rmccue (approved):
Classic.

roborourke:
Cannot wait to get Travis set up on these repos. Will all become immediately apparent if Local Server can't start.

PR #530, opened 2021-12-05
Refresh X-Ray docs for timeline view

rmccue:
See https://github.com/humanmade/vantage/pull/673


roborourke:
Will wait for an update here when the dependent change to the dashboard is released.

rmccue:
We shipped to /next/ so we should get this onto the nightly docs I think, and we can backport to the stable docs when we ship to / ?

roborourke:
/next/ still isn't something we link to even in the nightly docs though we could maybe do some filtering when we build the docs site to replace it.

Will review anyway.

docs/dashboard/x-ray.md
```diff
@@ -49,38 +49,68 @@ http.clientip = 1.2.3.4
...
 ## Understanding an XRay trace
 
-Each HTTP request is called an "XRay trace" and provides many useful pieces of debugging information.
+Each HTTP request generates an "X-Ray trace" and provides many useful pieces of debugging information.
+
+### Overview
+
+![](../assets/xray-overview.png)
+
+The overview table provides quick information about the request, including requested URL, total duration, response code, and external API calls to the database and remote servers.
```
roborourke, line 58: You mentioned this was response time rather than total duration now iirc?

rmccue:
I don't think we need to link out; I mean just that I want the docs in place for anyone who does try it out. :)

docs/dashboard/x-ray.md
```diff
@@ -49,38 +49,68 @@ http.clientip = 1.2.3.4
...
+![Example request data](../assets/xray-request.png)
+
+The request tab displays data sent to and served from the server, including the client IP, request and response headers, and any other statistics generated by the request. This can assist in debugging unexpected behaviour by replicating the conditions, or finding unexpected header values.
+
+Some information is redacted from the request data, for privacy reasons, including user passwords. You can redact additional information by filtering `aws_xray.redact_metadata_keys` or `aws_xray.redact_metadata`:
+
+```php
+// Redact keys from the metadata:
+add_filter( 'aws_xray.redact_metadata_keys', function ( $keys ) {
+    $keys['$_POST'][] = 'rcp_user_pass';
```
roborourke, line 83: This is a bit hard to understand, more so with the removal of what the metadata keys are e.g. `$_POST`, `$_GET` and `$_COOKIE`, `$_SERVER`.

When would I use redact metadata over redact metadata keys?

roborourke (commented):
I think we're losing a bit of info, and one of the changes might be slightly incorrect but otherwise all good.

docs/dashboard/x-ray.md
rmccue: That was in the table view, but @joehoyle I guess we need to update the docs to reflect that change too?

docs/dashboard/x-ray.md
rmccue: I asked myself the same question; I think basically if you don't know the keys ahead of time

roborourke: Right but we know some basics, I think this PR would be a good opportunity to expand on these filter docs but we can also add it later.

docs/dashboard/x-ray.md
roborourke:
```suggestion
The overview table provides quick information about the request, including requested URL, response time, response code, and external API calls to the database and remote servers.
```

docs/dashboard/x-ray.md
roborourke: @rmccue do you have time to make these updates or do you need me to complete it?

roborourke (approved):

PR #537, opened 2022-01-05
Correction to the configuration documentation

robindevitt:
By default Batcache is not on local environments. For developing cache customizations locally, you may wish to enable Batcache for local environments.

The example in the documentation is `"batcache": false`, where it should be `"batcache": true`.


shadyvb (approved):

robindevitt:
Closed to restart with the correct branch

Issue #544, opened 2022-01-18
Elastic requests debugging can result in empty messages.


shadyvb:
One support request identified an error message that doesn't include relevant information. The message is supposed to debug when a request to Elastic has failed, and include relevant error code and message to enable developers to further debug the issue and understand the failure reason.

The error is logged at https://github.com/humanmade/altis-cloud/blob/9.0.0/inc/namespace.php#L379.

The problem is that the `$response` can sometimes be a non-error, yet not have a non-2xx status codes, or not be an array at all, so we need to account for all possible cases to allow developers to debug such cases.

joehoyle:
Closing as wontfix / toonicetohave

PR #545, opened 2022-01-18
Improve logging failed Elastic requests

shadyvb:
Ref #544

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -379,13 +379,22 @@ function log_elasticsearch_request_errors( $response, string $context, string $c
 	$request_response_code = (int) wp_remote_retrieve_response_code( $response );
 	$is_valid_res = ( $request_response_code >= 200 && $request_response_code <= 299 );
 
+	if ( $is_valid_res ) {
+		return;
+	}
+
 	if ( is_wp_error( $response ) ) {
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		trigger_error( sprintf( 'Error in ElasticSearch request: %s (%s)', $response->get_error_message(), $response->get_error_code() ), E_USER_WARNING );
+	} elseif ( ! $request_response_code ) {
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		trigger_error( sprintf( 'Unknown error in ElasticSearch request: %s', wp_json_encode( $response ) ), E_USER_WARNING );
 	} elseif ( ! $is_valid_res ) {
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		trigger_error( sprintf( 'Error in ElasticSearch request: %s (%s)', wp_remote_retrieve_body( $response ), $request_response_code ), E_USER_WARNING );
+		trigger_error( sprintf( 'Error in ElasticSearch request: %s (%d: %s)', wp_remote_retrieve_body( $response ), $request_response_code, wp_remote_retrieve_response_message( $response ) ), E_USER_WARNING );
 	}
+
+	do_action( 'altis.cloud.elastic.failed_request', $response, $context, $class, $parsed_args, $url );
 }
 
 /**


inc/namespace.php
```diff
@@ -379,13 +379,22 @@ function log_elasticsearch_request_errors( $response, string $context, string $c
...
+	if ( $is_valid_res ) {
+		return;
+	}
+
 	if ( is_wp_error( $response ) ) {
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		trigger_error( sprintf( 'Error in ElasticSearch request: %s (%s)', $response->get_error_message(), $response->get_error_code() ), E_USER_WARNING );
+	} elseif ( ! $request_response_code ) {
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		trigger_error( sprintf( 'Unknown error in ElasticSearch request: %s', wp_json_encode( $response ) ), E_USER_WARNING );
```
shadyvb, line 391: Not entirely sure what's the best way to go about this, any better ideas ? @roborourke

shadyvb:
[ ] Still need to document the filter and make it visible in docs.

inc/namespace.php
roborourke: Ideally need to take some samples of the outputs provided at this stage. I'm not sure why some would be empty but perhaps the response code and headers would be useful to see 🤔

shadyvb: This is a sample of the output from the actual client implementation:

```
Warning[18-Jan-2022 11:35:06 UTC] PHP Warning: Unexpected error in ElasticSearch request: {"headers":{},"body":"","response":{"code":false,"message":""},"cookies":[],"filename":null,"http_response":{"data":null,"headers":null,"status":null}}
```

roborourke: oh right, so the request itself is empty... regarding indexing I've seen this happen because in the `prepare_document()` function of an `Indexable` subclass it can return `false` if the post isn't found for instance.

I only saw this when a site had an unclosed `switch_to_blog()` happening in some post content output. Note that the `the_content` filter is run when processing posts for indexing to get the full content.

shadyvb: It's rather the response that's empty, not the request.

roborourke: Is it not the full query? That object has `headers` and `body` properties then also `response` and `http_response` properties. 🤦‍♂️  - just remembered as I'm writing this that's how WP structures the response data.

shadyvb:
Quick update from further debugging:

> We found what appears to be the offending code: see split_large_ep_request. This code is responsible for causing the loss of useful log data.

> Added lower level custom logging which should correctly log failed requests which were split and logging bypassed by `split_large_ep_request`:
> 
> ```
> 
> add_action( 'http_api_debug', __NAMESPACE__ . '\\enhanced_es_debug', 10, 5 );
> 
> /**
> * Do an error log of failed ES requests
> * Lower level than inbuilt altis one, as it is able to catch any request made to ES via wp_remote_{x}
> *
> * @param array|WP_Error $response
> * @param string $context
> * @param string $class
> * @param array $parsed_args
> * @param string $url
> * @return void
> */
> function enhanced_es_debug( $response, $context, $class, $parsed_args, $url ) {
> 	if ( $context !== 'response' ) {
> 		return;
> 	}
> 
> 	// phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url
> 	$host = parse_url( $url, PHP_URL_HOST );
> 
> 	if ( ELASTICSEARCH_HOST !== $host ) {
> 		return;
> 	}
> 
> 	$request_response_body = wp_remote_retrieve_body( $response );
> 	$request_response_code = (int) wp_remote_retrieve_response_code( $response );
> 	$is_valid_res = ( $request_response_code >= 200 && $request_response_code <= 299 );
> 	$type = $parsed_args['method'] ?? 'GET?';
> 
> 	// Backup check for errors, sometimes the response is ok but the query
> 	// response JSON contains errors.
> 	$has_errors = strpos( $request_response_body, '"errors":true' ) !== false;
> 
> 	if ( is_wp_error( $response ) ) {
> 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
> 		trigger_error( sprintf( 'Error in ElasticPress request: WP_Error %s %s (%s) %s', $type, $response->get_error_message(), $response->get_error_code(), $url ), E_USER_WARNING );
> 	} elseif ( ! $is_valid_res || $has_errors ) {
> 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
> 		trigger_error( sprintf( 'Error in ElasticPress request: %s %s (%s) %s', $type, $request_response_body, $request_response_code, $url ), E_USER_WARNING );
> 	}
> }
> ```

joehoyle:
Going to close this as stale. Can re-open if needed.

PR #564, opened 2022-02-01
Add note about https redirection

owaincuvelier:
I think my wording is a bit messy but really wanted to jot something down and open a PR for it before I forget. If you have a better one liner for HTTPs redirect please share!

### DIFF
--- a/docs/https.md
+++ b/docs/https.md
@@ -14,4 +14,4 @@ HTTPS on Cloud requires user-agents support SNI (Server Name Indication). The fo
 
 # HTTPS Redirects
 
-It's common to force all traffic to route via HTTPS. To turn on forced HTTPS Redirection, [submit a support request](docs.altis-dxp.com/guides/getting-help-with-altis/) requesting that HTTPS Redirection be enabled. This is set at the CDN layer.
+It's common to force all traffic to route via HTTPS. To turn on forced HTTPS Redirection, [submit a support request](https://docs.altis-dxp.com/guides/getting-help-with-altis/) requesting that HTTPS Redirection be enabled. This is set at the CDN layer.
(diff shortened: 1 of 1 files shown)


shadyvb (approved):

docs/https.md
```diff
@@ -2,12 +2,16 @@
...
 
 HTTPS on Cloud requires user-agents support SNI (Server Name Indication). The following protocols are supported:
 
 - TLS 1.2
 - TLS 1.1
 - TLS 1.0
+
+# HTTPS Redirects
+
+It's common to force all traffic to route via HTTPS. To turn on forced HTTPS Redirection, [submit a support request](docs.altis-dxp.com/guides/getting-help-with-altis/) requesting that HTTPS Redirection be enabled. This is set at the CDN layer.
```
roborourke, line 17: Not sure if the link works without https://

roborourke:
@owaincuvelier it'll be beneficial to get into the habit of switching to the nightly docs version so your PRs are against the master branch rather than v9 etc...

docs/https.md
roborourke:
```suggestion
It's common to force all traffic to route via HTTPS. To turn on forced HTTPS Redirection, [submit a support request](https://docs.altis-dxp.com/guides/getting-help-with-altis/) requesting that HTTPS Redirection be enabled. This is set at the CDN layer.
```

owaincuvelier:
@roborourke yeah fair enough. I think I did forget. Nightly is just master, right?

roborourke:
Yep, it's not the end of the world, but ideally better to backport from master back to older versions but we can backport from here to master in this case.

owaincuvelier:
@roborourke that's fine! My bad!

PR #578, opened 2022-02-23
Document branch variables

owaincuvelier:
Adding a specific line about branch variables. 

I wanted to add a section on all the variables but couldnt think of good examples to illustrate their use with.

### DIFF
--- a/docs/build-scripts/README.md
+++ b/docs/build-scripts/README.md
@@ -127,6 +127,12 @@ console.log( 'Starting build...' )
 
 If you are using Composer dependencies from private repositories, [additional configuration may be required](./private-dependencies.md).
 
+### Targetting environments
+
+To create conditional build instructions dependant on environment, you can use the available `$BRANCH` variable. 
+
+`$BRANCH` and `$BRANCH_SLUG`, the latter being a sanitized version of the former.
+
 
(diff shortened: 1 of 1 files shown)


roborourke (commented):
Thinking it'd be good to review all available environment variables on the build container in case there are other things that'd be useful and document the available environment variables rather than just mentioning these.

At the very least this should show examples of what `$BRANCH` and `$BRANCH_SLUG` might actually look like relative to the GitHub branch.

roborourke:
Sorry, misread your comment - so agree on a section on all the variables, at least that are relevant to developers. I don't think we need to suggest examples for use-cases, just sample values.

roborourke:
@owaincuvelier are you able to finish this off based on my review or shall I get it assigned to someone and planned in?

owaincuvelier:
@roborourke I'll finish this off - but I did get a report that this wasn't actually working. Let me double check an ill update here shortly

owaincuvelier:
@roborourke yeah - so they weren't working, as it turns out. We've created: https://github.com/humanmade/altis-cloud/issues/591 

Lets close this PR - I've added a line for the acceptance criteria to document the variables. Sound good?

PR #601, opened 2022-03-25
Improve general documentation about Altis Cloud

rmccue:
* Adds an Architecture page and diagram
* Fleshes out many of the pages about components
* Clarifies that disabling components will void your warranty


docs/architecture.md
```diff
@@ -0,0 +1,94 @@
...
+Database servers will automatically scale based on load, adding and removing additional reader instances as needed (known as "horizontally scaling"). Reader instances are automatically distributed across availability zones, and can be promoted to writers for instant disaster recovery.
+
+Database servers can also be allocated more resources (known as "vertical scaling"). The Altis team manages your database instance type based on your typical load patterns and your environment plan. On higher tiers, your server infrastructure can also be allocated a database proxy, which acts as a load balancer for database connections, improving performance.
+
+
+## Object caching servers
+
+Each environment contains at least one Redis server, used for persistent object caching. This server acts as a key-value cache store, offloading expensive queries from the database.
+
+Object cache servers automatically scale based on load, just like the database servers. Altis automatically handles horizontally scaling these servers based on load, and the Altis team manages your instance type.
```
joehoyle, line 85: As noted, this is not the case, we don't horizontally scale, we have a fail-over.

Also, I thikn we should be mention it's the WP Object Cache specifically that's using the "object caching servers"

docs/architecture.md
```diff
@@ -0,0 +1,94 @@
...
+
+Running a web application like Altis requires running multiple infrastructure components. Altis uses dedicated, specialized services for each component, rather than handling each service on one machine. This provides better performance characteristics by ensuring services have the best resources for their need, and allows each service to scale independently.
+
+Every client and every environment is fully segregated from each other using logical controls at the network level, backed by AWS security controls.
+
+Altis Cloud is designed to be highly-available, with redundant servers split across multiple availability zones, which are physically-independent data centres. Each set of availability zones are located in the [same physical region](./origin-locations.md). This ensures resiliency and redundancy of your data and application.
+
+(Note that full HA is not included with all environment tiers; consult your plan information. High availability is also only available on production environments.)
+
+![Architecture diagram](assets/architecture.svg)
```
joehoyle, line 35: FYI each customer does not have a VPC, which this diagram seems to suggest.

docs/database.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Multi-Server Configuration
+
+Unlike many WordPress hosts, Altis Cloud environments use a multi-server configuration for the database, with fully dedicated servers for your database. Servers are categorised into two general groups:
+
+* **Primary**: The primary server is the authoritative server storing all data.
+* **Replica:** The replica servers store copies of the data, continually synchronised from the primary.
+
+All database writes (`INSERT`, `UPDATE`, `DELETE`, etc) are sent directly to the primary server, while database reads (`SELECT`) are sent to the replicas. This allows offloading most of the database load from the primary.
```
joehoyle, line 15: `SELECT`s can be sent to either writers or readers, might not be important though.

docs/database.md
```diff
@@ -0,0 +1,66 @@
...
+For customers who regularly run larger numbers of replicas, an additional [Proxy server](https://aws.amazon.com/rds/proxy/) is used to pool connections and more evenly distribute load across the server pool. The Altis team manages the provisioning of Proxy servers on your behalf.
+
+
+## Autoscaling and Performance
+
+As load on your site increases, the Altis Cloud infrastructure will automatically start new replica servers as necessary to cope with database load. Autoscaling occurs primarily based on CPU usage of your database servers, with other factors incorporated.
+
+In order for autoscaling to work well and for your site to remain performant, **writes should be avoided on frontend requests**. While replicas can easily be created, in typical configurations only a single primary server exists. Writes to the database cause more load upon the database server, and since the primary cannot easily be automatically scaled, large numbers of writes may cause your database to be overloaded.
+
+(As most frontend requests are also [cached at the CDN](./page-caching.md), writes may not be consistent.)
```
joehoyle, line 30: I get what you are saying, but I think calling writes not consistent might imply that the database layer is not consistent.

docs/database.md
```diff
@@ -0,0 +1,66 @@
...
+
+Sites expecting large numbers of logged in users or many database writes (such as ecommerce sites) may need increased provisioning for the primary database server, changes to the autoscaling behaviour, or a multi-primary setup. Altis Cloud environments can accomodate this behaviour upon arrangement; please note that this may incur extra cost.
+
+
+## Configuration
+
+### Versions
+
+Altis provides support for MySQL 5.6 or 5.7.
```
joehoyle, line 39: I think we might be 5.7 only now? I.e. if someone asked for 5.6, would we let them?

docs/nginx-configuration.md
```diff
@@ -1,8 +1,33 @@
...
+
+
+## Providing Configuration
+
+Nginx will look for an additional file in your project repository at `.config/nginx-additions.conf` and will load it into the `server {}` context of the main configuration. You can also use filename suffixes to split your configuration in complex cases.
+
+Internally, the Altis Nginx configuration looks like:
+
+```
+server {
```
joehoyle, line 15: We also recently added for https://github.com/humanmade/docker-wordpress-nginx/pull/62 `nginx-http-additions` so you can do things like use `map` etc.

docs/object-cache.md
```diff
@@ -9,21 +9,22 @@ Altis provides a low-level integration with WordPress's object caching system ou
...
+
+
+## Infrastructure
+
+Unlike many WordPress hosts, Altis Cloud environments use a multi-server configuration for the object cache, with fully dedicated servers for your object cache. Servers are categorised into two general groups:
+
+* **Primary**: The primary server is the authoritative server storing all data.
+* **Replica:** The replica servers store copies of the data, continually synchronised from the primary.
+
+Just like with your [database servers](./database.md), object cache writes are sent directly to the primary server, while reads are sent to the replicas. This allows offloading most of the load from the primary.
```
joehoyle, line 22: As noted, this is wrong.

joehoyle (changes requested):

docs/architecture.md
rmccue: It's a semi-intentional fudge, not really relevant (and we're moving towards that anyway)

docs/database.md
rmccue: D'oh, didn't think about the term "consistent" there, that was silly

docs/database.md
rmccue: True.

docs/nginx-configuration.md
```diff
@@ -1,8 +1,49 @@
...
+**Note:** Custom configuration will only apply to traffic served by the web container, which does not include media or tachyon (see the [architecture diagram for more information](./architecture.md)). Specifically, any URLs beginning with `/uploads/` or `/tachyon/` are routed directly to [S3](./s3-storage.md) and [Tachyon](docs://media/dynamic-images.md) directly.
+
+
+## Server Configuration
+
+Custom configuration can be supplied in Nginx either within the `server {}` context, or the `http {}` context.
+
+Nginx loads the following custom files from your project repository:
+
+* `.config/nginx-http-additions.conf`: Loaded into the `server` context
```
joehoyle, line 14:
```suggestion
* `.config/nginx-http-additions.conf`: Loaded into the `http` context
```

joehoyle (approved):
Need to fix the code suggestion, but then good to go.

PR #614, opened 2022-05-03
Correct the reference to `mysqli` so database error code appears correctly in QM

johnbillion:
When a database query results in an error, the generic `qmdb` error code is shown in Query Monitor instead of the actual error code from MySQL. This is because the reference to `mysqli` isn't namespaced correctly.

Before:

After:


roborourke (approved):
derp

PR #615, opened 2022-05-05
Update origin-locations.md

johnbillion:


roborourke (approved):

roborourke:
Thanks John, was this change based on some feedback or just felt odd that some were cities and some countries?

johnbillion:
Yeah I needed to confirm that we run out of a data centre in Germany and it was weird seeing a mixture of country names and cities.

PR #640, opened 2022-05-31
Add more details and examples to nginx config

rmccue:
Adds further docs about using nginx for access controls, including IP-based restrictions.


roborourke:
Ready for review here @rmccue ?

rmccue:
Need to test it 😅

roborourke:
@robindevitt would you be able to test out these docs Ryan has written up?

roborourke:
@rmccue can you test this? Been hanging around a while

PR #665, opened 2022-08-23
Bootstrap XRay if Excimer is available

kovshenin:
Currently XRay is bootstrapped only if XHProf is available, which we no longer build for PHP 8.0 and above. The upstream XRay plugin supports the newer Excimer profiler that is PHP 8.0+ compatible. This change bootstraps the plugin if either of the two profilers is available.


roborourke:
Whoops

roborourke (approved):

jerico (approved):

mikelittle (approved):
Looks good to me

PR #679, opened 2022-09-13
Ensure `SAVEQUERIES` is defined before checking its value

robindevitt:
Originally identified in - https://github.com/humanmade/altis-documentation/pull/425

---

Running `composer dev-tools codecept -p  packages/documentation/tests run -vvv` I was able to identify the first error:
```php
PHP Fatal error:  Uncaught Error: Undefined constant "Altis\Cloud\SAVEQUERIES" in /usr/src/app/packages/cloud/inc/class-db.php:63
```

After applying the fix for it, the following error came up:
```php
PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in /usr/src/app/packages/cloud/inc/class-db.php:69
```
Applying the two fixes above allows `composer dev-tools codecept -p  packages/documentation/tests run -vvv` to run smoothly.

### DIFF
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -12,8 +12,6 @@ use LudicrousDB;
 use QM_Backtrace;
 use WP_Error;
 
-const SAVEQUERIES = true;
-
 /**
  * Ludicrous DB integration class.
  *
@@ -62,11 +60,11 @@ class DB extends LudicrousDB {
 		$end = microtime( true );
 		$this->time_spent += $end - $start;
 
-		if ( ! $has_qm || ! SAVEQUERIES ) {
+		if ( ! is_array( $this->queries ) || ! $has_qm || ! defined( 'SAVEQUERIES' ) || ! SAVEQUERIES ) {
 			return $result;
 		}
 
-		$i = count( (array) $this->queries ) - 1;
+		$i = count( $this->queries ) - 1;
 		$this->queries[ $i ]['trace'] = new QM_Backtrace( [
 			'ignore_frames' => 1,
 		] );
--- a/inc/class-db.php
+++ b/inc/class-db.php
@@ -12,6 +12,8 @@ use LudicrousDB;
 use QM_Backtrace;
 use WP_Error;
 
+const SAVEQUERIES = true;
+
 /**
  * Ludicrous DB integration class.
  *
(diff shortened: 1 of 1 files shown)


robindevitt:
Noting that I do find the tests to take a long time to run, sometimes they fail but purely from a time stand. Re-running the task usually goes through again.

robindevitt:
This only info around `SAVEQUERIES` I could find was here : https://docs.altis-dxp.com/dev-tools/testing-with-codeception/#wpqueries

inc/class-db.php
```diff
@@ -12,6 +12,8 @@
 use QM_Backtrace;
 use WP_Error;
 
+const SAVEQUERIES = true;
```
roborourke, line 15: Where is `SAVEQUERIES` normally defined? Note that the error message isn't saying that the constant `SAVEQUERIES` _should_ exist in this namespace - normally it would be defined using `define()` somewhere globally. We need to understand and account for the situation where that isn't the case rather than just define it here.

We wouldn't necessarily want `SAVEQUERIES` to be `true` in production either for example.

inc/class-db.php
```diff
@@ -64,7 +66,7 @@ public function query( $query ) {
 			return $result;
 		}
 
-		$i = count( $this->queries ) - 1;
+		$i = count( (array) $this->queries ) - 1;
```
roborourke, line 69: The error message says that `$this->queries` is actually `null`. We should understand why it can be `null` in some cases and most likely return early rather than just cast it as an array here. `null` is not the same as `[ null ]` which will have a `count()` of 1.

roborourke (changes requested):
It's on the right path but rather than sweep the errors under the rug so to speak, we should understand why they happen and account for that.

roborourke:
@robindevitt https://wordpress.org/support/article/debugging-in-wordpress/#savequeries

So really the check here should be looking to see if save queries is defined before checking its value rather than defining it.

inc/class-db.php
kovshenin: SAVEQUERIES is usually defined in wp-config.php, sometimes forced by tools like Query Montior in db.php. Anything later than that would be too late.

kovshenin (approved):
Change looks good.

kovshenin:
> So really the check here should be looking to see if save queries is defined before checking its value rather than defining it.

Agreed. Defining it in our own namespace won't do anything anyway, except possibly produce side-effects where some of our code thinks saving queries is on, while in reality is off. WordPress and QM define and check the constant in the global scope.

roborourke:
@robindevitt sorry to take over this one, needed to get the change in asap, let me know if you want to talk through the alternative solution when you're back.

robindevitt:
@roborourke looks good, thanks for moving it forward 👍🏼

PR #697, opened 2022-10-24
geo-targeting is enabled by defualt

owaincuvelier:


johnbillion:
The reason I brought this up is because on Wüest we're investigating lower than expected cache hit rates and we're only using the `Cloudfront-Viewer-Country` header on requests to the root URL (`/`). Ideally we'd only vary the cache key based on this header on the root URL and disregard it otherwise.

owaincuvelier:
@rmccue lets merge this is in anyway and we can revert the docs once the infra changes are in place

rmccue:
@owaincuvelier I agree on updating the docs to reflect the current situation 👍

roborourke (approved):

PR #698, opened 2022-10-25
Add documentation on Performance Expectations

joehoyle:


docs/performance-expectations.md
```diff
@@ -0,0 +1,17 @@
+# Performance Expectations
+
+## CDN Response Times
+
+Altis Cloud serves [static files](./static-file-caching.md) and the [page cache](./page-caching.md) from the CDN. In cases when static files and pages are already cached on the CDN, response times for client requests are typically between 10 miliseconds and 30 miliseconds. This can depend on the quality of connection from the client. See [CDN](./cdn/README.md) for details on Edge Cache locations.
```
rmccue, line 5:
```suggestion
Altis Cloud serves [static files](./static-file-caching.md) and the [page cache](./page-caching.md) from the CDN. In cases when static files and pages are already cached on the CDN, response times for client requests are typically between 10 milliseconds and 30 milliseconds.

Response times can depend on the quality of connection from the client. See [CDN](./cdn/README.md) for details on Edge Cache locations.
```

docs/performance-expectations.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Origin Cache Response Times
+
+The origin cache (also called Batcache) is documented in [page cache](./page-caching.md). Response times for Batcache HIT requests are typically between 30 miliseconds and 60 miliseconds.
```
rmccue, line 9:
```suggestion
The origin cache (also called Batcache) is documented in [page cache](./page-caching.md). Response times for Batcache HIT requests are typically between 30 milliseconds and 60 milliseconds.
```

docs/performance-expectations.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Origin Response Times
+
+Origin response times when PHP / WordPress dynmically generate a page can vary dignificantly depending on the site's implementation (plugins used, WordPress theme, database queries etc). We're provided the following guidelines for page generation time to help teams better understand the typical "base" response times. These figures may not be illustritive of all projects given how many factors dictate the performance of WordPress based sites.
```
rmccue, line 13:
```suggestion
Origin response times when PHP / WordPress dynamically generates a page can vary significantly depending on the site's implementation, including which plugins are used, the active WordPress theme, how many database queries are used, etc. The included [development tools](docs://dev-tools/) help with visibility of these statistics.

We're provided the following guidelines for page generation time to help teams better understand the typical "base" response times, based on our real-world experience.

These figures may not be illustrative of all projects, but can provide a guideline for optimisation.
```

docs/performance-expectations.md
```diff
@@ -0,0 +1,17 @@
...
+
+|Page Generation Type|Average (miliseconds)|90th Percentile (miliseconds)|
+|--|--|--|
+|New WordPress Install + `Twenty Twenty Two` theme homepage.|150|450|
```
rmccue, line 17:
```suggestion
Page Generation Type | Average (milliseconds) | 90th Percentile (milliseconds)
-- | -- | --
New WordPress Install + `Twenty Twenty Two` theme homepage. | 150 | 450
```

rmccue (changes requested):

docs/performance-expectations.md
joehoyle: I'll take your word for it!

rmccue (approved):

kovshenin (approved):

joehoyle:
@johnbillion the intention is to provide baseline metrics so people deploying on Altis Cloud can understand the delta between their "site additions" and what a blank slate vanilla WordPress is likely to perform at. It also provides a baseline so they understand what minimum numbers are viable (if for example, they expect to optimize their code to get things responding in 50ms).

I think this doc could be fleshed out in the future though, with some more example setups. I don't think we should be blocking merging it now though.

PR #721, opened 2023-01-16
Update Composer version

johnbillion:
According to `composer --version` inside a `.build-script` file, Composer is now at version 2.4.4 on the cloud deployment infrastructure.


mikelittle (approved):
Thanks, John

PR #731, opened 2023-03-17
Add support for Afterburner

### LINKED ISSUE #748: Add support for Afterburner
When Afterburner is enabled for the application, the legacy cache drop-in should not be loaded.

The drop-in, `object-cache.php`, is normally loaded early in the WordPress boot-up process.

### Acceptance criteria

* [ ] Legacy drop-in is not loaded if the Afterburner extension is enabled in Local Server
* [ ] Legacy drop-in is not loaded if the Afterburner extension is enabled in Altis Cloud

joehoyle:
We don't have a solution for Sessions yet, we could still use predis for that if we wanted.

See https://github.com/humanmade/product/issues/567

Fixes #748

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -193,6 +193,10 @@ function load_platform( $wp_debug_enabled ) {
 		load_session_handler();
 	}
 
+	if ( $config['ludicrousdb'] ) {
+		load_db();
+	}
+
 	if ( $config['cdn-media-purge'] ) {
 		load_cdn_media_purge();
 	}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -425,12 +425,6 @@ function load_object_cache_redis() {
  * Configure Afterburner's Redis connection for the object cache.
  */
 function load_object_cache_afterburner() {
-	if ( defined( 'REDIS_AUTH' ) && defined( 'REDIS_HOST' ) && defined( 'REDIS_PORT' ) ) {
-		$secure = defined( 'REDIS_SECURE' ) ? REDIS_SECURE : true;
-		ini_set( 'afterburner.redis_server_info', sprintf( 'redis%s://:%s@%s:%d', $secure ? 's' : '', REDIS_AUTH, REDIS_HOST, REDIS_PORT ) );
-	} else {
-		trigger_error( 'REDIS constants not defined.', E_USER_ERROR );
-	}
 	require __DIR__ . '/alloptions_fix/namespace.php';
 	Alloptions_Fix\bootstrap();
 	// cache must be initted once it's included, else we'll get a fatal.
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -191,6 +191,8 @@ function load_platform( $wp_debug_enabled ) {
 	} elseif ( $config['redis'] ) {
 		load_object_cache_redis();
 		load_session_handler();
+	} else {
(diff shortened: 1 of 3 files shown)


rmccue:
> We don't have a solution for Sessions yet, we could still use predis for that if we wanted.

I think that may be required for WooCommerce support?

rmccue:
Split sessions to https://github.com/humanmade/afterburner/issues/14

inc/namespace.php
```diff
@@ -454,6 +420,22 @@ function load_object_cache_redis() {
 	wp_cache_init();
 }
 
+/**
+ * Configure Afterburner's Redis connection for the object cache.
+ */
+function load_object_cache_afterburner() {
+	if ( defined( 'REDIS_AUTH' ) && defined( 'REDIS_HOST' ) && defined( 'REDIS_PORT' ) ) {
+		$secure = defined( 'REDIS_SECURE' ) ? REDIS_SECURE : true;
+		ini_set( 'afterburner.redis_server_info', sprintf( 'redis%s://:%s@%s:%d', $secure ? 's' : '', REDIS_AUTH, REDIS_HOST, REDIS_PORT ) );
```
kovshenin, line 429: How safe is changing this mid-request?

joehoyle: Basically this needs to be set before `wp_cache_init()` is called. If `wp_cache_init()` is called before a valid redis server is provided, it will also re-try with the new details the next time wp_cache_init() is called so it should be safe 👍

kovshenin: Ok, either way we set this setting (and others) in docker-entrypoint when running php-fpm so I don't think it's necessary to set again here.

If we do want to support changing the connection string like that, I think we should detect that change in wp_cache_init and explicitly close the previous main thread and pubsub connections, otherwise setting the connection string is a shot in the dark, which will only work if/when the previous connection fails.

joehoyle: Ah cool, if you are setting already via `php -fpm -d` (or ini) then that's fine, we can remove.

inc/namespace.php
```diff
@@ -181,21 +181,15 @@ function load_platform( $wp_debug_enabled ) {
...
-
-	// Session support.
-	if ( $config['redis'] ) {
 		load_session_handler();
-	} else {
-		disable_sessions();
-	}
-
-	if ( $config['ludicrousdb'] ) {
-		load_db();
```
kovshenin, line 198: Doesn't seem like this was removed intentionally @joehoyle?

inc/namespace.php
```diff
@@ -181,21 +182,15 @@ function load_platform( $wp_debug_enabled ) {
...
```
kovshenin, line 198: Doesn't seem like this was removed intentionally @joehoyle?

inc/namespace.php
```diff
@@ -181,17 +182,15 @@ function load_platform( $wp_debug_enabled ) {
...
+		disable_sessions();
 	} elseif ( $config['redis'] ) {
 		load_object_cache_redis();
-	}
-
-	// Session support.
-	if ( $config['redis'] ) {
 		load_session_handler();
-	} else {
-		disable_sessions();
```
kovshenin, line 194: Should we still disable sessions if Afterburner/Redis/Memcache are not enabled?

PR #760, opened 2023-04-28
Add a WP Cache session handler

kovshenin:
Adds support for PHP sessions with our Afterburner implementation of the WordPress object cache.

### DIFF
--- a/inc/session_handler/class-wp-cache-session-handler.php
+++ b/inc/session_handler/class-wp-cache-session-handler.php
@@ -72,9 +72,9 @@ class WP_Cache_Session_Handler implements SessionHandlerInterface {
 	 *
 	 * @param string $id Session id
 	 *
-	 * @return string|false
+	 * @return string
 	 */
-	public function read( string $id ) : string | false {
+	public function read( string $id ) : string {
 		$data = wp_cache_get( $id, 'sessions' );
 		if ( ! $data ) {
 			return '';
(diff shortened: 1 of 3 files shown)


inc/session_handler/class-wp-cache-session-handler.php
```diff
@@ -0,0 +1,104 @@
...
+	/**
+	 * Read session data.
+	 *
+	 * @param string $id
+	 *
+	 * @return string|false
+	 */
+	public function read( string $id ) : string | false {
+		$data = wp_cache_get( $id, 'sessions' );
+		if ( ! $data ) {
```
joehoyle, line 74: This may want to use `$found`? or I guess if it's only ever a string, it should be safe enough?

joehoyle:
Very cool! I think the linter issues need fixing up.

inc/session_handler/class-wp-cache-session-handler.php
kovshenin: Yeah, it throws a warning if you return anything that's not a string (probably because `open()` succeeded), so even for a session that was not found we have to return an empty string. We know `wp_cache_get()` will return `false` if the value doesn't exist, because `$data` is a `string` in `write()` but that still should be returned as an empty string to avoid the warning.

I believe this is because `open()` is typically responsible for determining whether the session exists, i.e. tries to open a file descriptor, so when a descriptor is open, `read()` is almost guaranteed to succeed. Maybe moving `wp_cache_get()` to `open()`, storing the value in a property and then returning it in `read()`? Maybe that's more semantically accurate, but that makes things more confusing, especially given that the Redis connection is abstracted anyway. Forcing a string is a good tradeoff I think.

joehoyle (approved):
Nice, looks good

PR #762, opened 2023-05-08
Update class-cli-command.php

johnbillion:
This corrects the command description when viewing the help in WP-CLI.

## To test

* Run `composer server cli -- help`


mikelittle:
Thanks for this.

PR #776, opened 2023-05-22
Add documentation for Afterburner.

mikelittle:
Mention it in the architecture page.
(Note modified lines have been reformatted to MD standard)

Fixes https://github.com/humanmade/product-dev/issues/1354

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -62,12 +62,17 @@ Three types of application containers are run within your environment:
 
 * **Cavalcade containers** handle [running scheduled and background tasks](./scheduled-tasks.md). These automatically scale based on the number of pending tasks, allowing huge numbers of tasks to be run simultaneously.
 
-Containers are treated as "throwaway" instances, which contain a copy of your codebase and Altis modules. No data is stored in application containers, allowing problematic containers to be removed and replaced at any time. Containers can also be added or removed dynamically, handling load during high traffic events (known as "horizontal scaling").
+Containers are treated as "throwaway" instances, which contain a copy of your codebase and Altis modules. No permanent
+data is stored in application containers, allowing problematic containers to be removed and replaced at any time.
+Containers can also be added or removed dynamically, handling load during high traffic events (known as "horizontal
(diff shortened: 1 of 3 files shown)


rmccue:
I feel like we should potentially make this a broader page about Afterburner? ie include the translations stuff too, and anything else we might add later.

docs/architecture.md
```diff
@@ -62,12 +62,17 @@ Three types of application containers are run within your environment:
...
+Containers are treated as "throwaway" instances, which contain a copy of your codebase and Altis modules. No permanent
+data is stored in application containers, allowing problematic containers to be removed and replaced at any time.
+Containers can also be added or removed dynamically, handling load during high traffic events (known as "horizontal
+scaling").
 
 Traffic is routed to containers through a load balancer. This load balancer ensures traffic is distributed across containers evenly, and ensures that containers being added or removed are seamlessly added to the pool. The load balancer also acts as the firewall and ingress point to your environment, with strict security controls at the network and application layers.
 
 Web containers automatically collect application performance monitoring (APM) data, which is available through the Altis Dashboard.
 
+The application containers can use our [Afterburner Cache](./afterburner-cache.md) which improves object cache
```
rmccue, line 74:
```suggestion
The application containers use our [Afterburner Cache](./afterburner-cache.md) which improves object cache
```

mikelittle:
That's why I named the page `afterburner-cache` so as to leave room for a general page about afterburner. But I can make the adjustment now if that makes sense.

docs/architecture.md
mikelittle: Is the plan that it will always be enabled? I put 'can' because it needs to be turned on at the moment.

rmccue: Yeah, that's just for the pilot program; it'll roll out to everyone with the major release.

rmccue:
I think a single page for all Afterburner stuff will make more sense than, say, one per feature

kovshenin (approved):

PR #794, opened 2023-07-20
Add a reference to PHP 8.2

mikelittle:


docs/specifications.md
```diff
@@ -7,7 +7,7 @@ order: 100
 
 The following versions are available, and the specific version in use will depend on your configuration.
 
-- PHP Version: 7.4, 8.0, 8.1
+- PHP Version: 7.4, 8.0, 8.1, 8.2 See [Altis compatibility chart](docs://guides/updating-php/#altis-compatibility-chart)
```
owaincuvelier, line 10: I think we should add an "experimental" tag to PHP 8.2 for now. Thought?

mikelittle: The linked compatibility chart notes that both 8.1 and 8.2 are experimental.

docs/specifications.md
```diff
@@ -7,7 +7,7 @@ order: 100
...
 - MySQL Version: 5.6, 5.7 (compatible; using AWS Aurora)
```
ferschubert-hm, line 11: Can we remove 5.6 version is no longer in use and has been fully deprecated also by AWS Aurora.

ferschubert-hm (changes requested):
Left a comment to remove deprecated MySQL version as well

docs/specifications.md
mikelittle: Done. Thanks

ferschubert-hm (approved):

PR #826, opened 2023-10-18
Update afterburner requirements

svandragt:
Added missing afterburner requirements.


mikelittle:
Thanks for this, but as v16 and v17 are no longer supported and therefore every supported version has the requirements, I'm going to close this PR.

PR #841, opened 2024-02-05
URL encode filenames in CDN Media Purge

joehoyle:
CloudFront will return an error from `createInvalidation` if the URL is not encoded. `tachyon_url()` already does this internally, so is not an issue on the tachyon_url.

### DIFF
--- a/inc/cloudfront_media_purge/namespace.php
+++ b/inc/cloudfront_media_purge/namespace.php
@@ -32,7 +32,7 @@ function purge_media_file_cache( int $media_id ) {
 	$upload_path_info = pathinfo( $upload_path );
 	$items            = [];
 	// Make sure the image filename is URL encoded, else CloudFront will not be able to purge it.
-	$items[]          = $upload_path_info['dirname'] . '/' . urlencode( $upload_path_info['filename'] ) . '*';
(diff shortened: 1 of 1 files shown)


mikelittle (approved):
Looks good, thanks Joe, but the first linter issue (extra space) could be fixed.

kovshenin (approved):

jerico (approved):

inc/cloudfront_media_purge/namespace.php
```diff
@@ -31,7 +31,8 @@ function purge_media_file_cache( int $media_id ) {
 	$upload_path      = wp_parse_url( $upload_url, PHP_URL_PATH );
 	$upload_path_info = pathinfo( $upload_path );
 	$items            = [];
-	$items[]          = $upload_path_info['dirname'] . '/' . $upload_path_info['filename'] . '*';
+	// Make sure the image filename is URL encoded, else CloudFront will not be able to purge it.
+	$items[]          = $upload_path_info['dirname'] . '/' . urlencode( $upload_path_info['filename'] ) . '*';
```
rmccue, line 35: Linter warning here; variables shouldn't be aligned any more, but even if they are, comments reset alignment.
```suggestion
	$items[] = $upload_path_info['dirname'] . '/' . urlencode( $upload_path_info['filename'] ) . '*';
```

rmccue (changes requested):

PR #859, opened 2024-03-01
Fix < php 8.0

joehoyle:
Union return type was only added in 8.0. So this fails the hm-linter checks


joehoyle:
ping @kovshenin to get the project green

mikelittle:
Is this necessary? Altis master doesn't support PHP < 8.0
Is the fix to configure the automatic lint to check against PHP >= 8.0

mikelittle (approved):
Looks good to me

PR #864, opened 2024-03-29
Disable PHP language files with Afterburner

mikelittle:
Prefer only `.mo` files for translations if the Afterburner extension is loaded.

References https://github.com/humanmade/product-dev/issues/1546

Before:
CleanShot 2024-03-29 at 16 56 36

After:
CleanShot 2024-03-29 at 17 07 07


mikelittle:
Closing this as it does not solve the problem.

PR #873, opened 2024-04-11
Fix Afterburner translations with WordPress 6.5.0

kovshenin:
Core 6.5.0+ introduced new translation classes and routines, removing the need for (and use of) the MO object, which Afterburner pomo relies on. This change brings back the old behavior when loading text domains.

See: https://github.com/humanmade/product-dev/issues/1546

### DIFF
--- a/inc/afterburner/namespace.php
+++ b/inc/afterburner/namespace.php
@@ -76,3 +76,4 @@ function pre_load_textdomain( $load, $domain, $mofile, $locale ) {
 	$wp_textdomain_registry->set( $domain, $locale, dirname( $mofile ) );
 	return true;
 }
+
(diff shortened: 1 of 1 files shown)


rmccue:
lgtm!

We did a bunch of testing here, with the final result being:

Screenshot 2024-04-15 at 15 51 11@2x

(core-php is core with `.l10n.php` file, core-mo is core with `.mo` file, afterburner-mo is Afterburner with `.mo` file, core-php-noopcache is core with `.l10n.php` file _and_ opcache disabled via `ini_set( 'opcache.enable', 0 );`)

tldr: Core has made some massive improvements! Afterburner is still faster except for `load_textdomain()` (just like before). 

Core also has a new trick where `load_textdomain()` doesn't do anything, and only lazy-loads the .mo on the first usage. By also introducing a proxy object between the translation calls and the actual files, it's able to avoid actually parsing a lot of files. We should implement both of these tricks in a future version of Afterburner.

mikelittle:
This looks good to me. Is there anything else to be done? Happy to approve it and get it into v19.

mikelittle (approved):
Looks good.

PR #874, opened 2024-04-22
Update dependabot for v19 release

mikelittle:


.github/dependabot.yml
```diff
@@ -52,4 +52,4 @@ updates:
   allow:
   - dependency-type: "direct"
   versioning-strategy: "increase"
-  target-branch: v14-branch
+  target-branch: v15-branch
```
kovshenin, line 55: No longer need 15 it seems: https://docs.altis-dxp.com/releases/

kovshenin (approved):

.github/dependabot.yml
mikelittle: True. I'll create another set of PRs to remove v15!

PR #896, opened 2024-09-20
Specify cookie cache overrides

rmccue:
We should clarify exactly how the cookie caching behaviour works, as this can potentially be confusing.


mikelittle (approved):
Looks good to me

mikelittle:
Merged it for ya.

PR #897, opened 2024-10-01
Reduce number of API calls to PutMetricData

joehoyle:
This should reduce the calls to 1 from 9.


joehoyle:
Tested locally with custom AWS credentials for the test account in the composer.json to confirm this shows up in CloudWatch.

jerico (approved):

PR #912, opened 2024-12-06
Disable loading column values in audit log by default

joehoyle:
Calculating the column values is very slow and mostly leads to timeouts on loading the page. Having the column values just improves the UI slightly by greying out values that are not in the DB. It doesn't stop the `<select>` from showing all filter options.

I added this as a filter incase anyone would still want this functionality.


rmccue (approved):
Not sure I really understand this, but sure.

PR #913, opened 2024-12-06
Flush Cavalcade Logs every job

joehoyle:
Currently we're buffering up to 1,000 log entries (with a max of 24 hours) for cavalcade logs which makes a significant delay in them appearing in the Altis Dashboard.

This doesn't change the cost of PutLogEvents which is based off event count, regardless of whether they are in separate API calls or not.

Fixes https://github.com/humanmade/product-dev/issues/1664

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -1079,9 +1079,12 @@ function is_cloud() : bool {
  *                          will not be created if it does not already exist.
  * @param string $log_stream Name of the log stream. This value is used as
  *                           provided.
+ * @param int $batch_size The amount of log items to buffer before flushing them.
+ *                        FLogs are automatically flushed / written out when the
+ *                        LoggerInterface is destructed. Defaults to 1000.
  * @return Psr\Log\LoggerInterface
  */
-function get_logger( string $log_group, string $log_stream ) : LoggerInterface {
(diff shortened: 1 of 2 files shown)


ferschubert-hm (approved):
LGTM

inc/namespace.php
```diff
@@ -1119,7 +1122,7 @@ function get_logger( string $log_group, string $log_stream ) : LoggerInterface {
 			Altis\get_environment_name() . '/' . $log_group,
 			$log_stream,
 			null, // log retention when creating a new group (we disable this).
-			1000, // how many logs to send in a single batch.
+			$batch_size, // how many logs to send in a single batch.
```
rmccue, line 1125: If there's a backlog of (eg) 20 logs and then this is set to 1, wouldn't that instruct it to send 20 requests to flush the existing ones? Feels to me like it would be better to force-flush when needed.

joehoyle: @rmccue there couldn't be a backlog of 20 logs if this is set to 1, batch === buffer size

rmccue: I mean if something had previously called `get_logger()` for the same group - or do we not dedupe?

joehoyle: No, there's no depublication

Issue #866, opened 2024-04-01
`esc_html` is not safe for exception handling in wp-redis


roborourke:
This is a bug inherited after merging the upstream changes to `wp-redis` to our fork.

Issues on the pantheon repo:

- https://github.com/pantheon-systems/wp-redis/issues/410
- https://github.com/pantheon-systems/wp-redis/issues/418

PR merged on upstream repo:

- https://github.com/pantheon-systems/wp-redis/pull/421

We need to merge the upstream changes again, release and update the cloud module.

If we're not maintaining any of our own changes to the upstream version any more, or if we only forked it to publish to packagist, I would suggest we change Altis to include the upstream as a dependency now that Pantheon have published it https://packagist.org/packages/pantheon-systems/wp-redis

After discussion, it looks like there is no longer any need to keep the HM fork.
---

## Acceptance Criteria

- [ ] Remove our forked version and change to upstream
- [ ] Test everything still works.
- [ ] Archive HM fork 

---

### Ready for Work Checklist

Is this ticket ready to be worked on? See
[the Play Book Definition of Ready](https://playbook.hmn.md/play/product/definition-of-ready/)

- [x] Is the title clear?
- [x] Is the description clear and detailed enough?
- [x] Are acceptance criteria listed?
- [x] Have any dependencies been identified? (Optional)
- [x] Have any documentation/playbook changes been identified? (Optional)
- [ ] Is an estimate or time box assigned?
- [x] Is a priority label assigned?
- [ ] Is this ticket added to an epic? (Optional)

### Completion Checklist

Is this ticket done? See
[the Play Book Definition of Done](https://playbook.hmn.md/play/product/definition-of-done-2/)

- [ ] Has the acceptance criteria been met?
- [ ] Is the documentation updated (including README)?
- [ ] Do any code/documentation changes meet project standards?
- [ ] Are automatic tests in place to verify the fix or new functionality?
- [ ] Or are manual tests documented (at least on this ticket)?
- [ ] Are any Playbook/Handbook pages updated?
- [ ] Has a new module release (patch/minor) been created/scheduled?
- [ ] Have the appropriate `backport` labels been added to the PR?

mikelittle:
I added to refine and must have labels. 
We need to check we are ok to switch to the upstream version, or, if we are changing it in some way, incorporate the bug fix.

mikelittle:
[...]

PR #928, opened 2025-01-09
Switch to upstream wp-redis

mikelittle:
Now that the Pantheon Systems original `wp-redis` is available from [packagist.org](https://packagist.org/packages/pantheon-systems/wp-redis), we can switch to using it.

Fixes https://github.com/humanmade/altis-cloud/issues/866


wisyhambolu (approved):
Looks good!

PR #929, opened 2025-01-24
Update Predis adaptor to 0.1.2

joehoyle:
Includes https://github.com/humanmade/wp-redis-predis-client/pull/14 which will show when cache flushes happen


rmccue (approved):
Do you want to backport this at all?

joehoyle:
Yes, we need this on a client with v20, so I'll backport to 21 & 21

Issue #712, opened 2022-12-04
Document how to allow blocked files


rmccue:
By default, certain types of requests are blocked, but you might want to allow these through. For example, all requests for files starting with `.` are blocked, but `/.well-known` is part of several important protocols that you might want to handle.

These can be handled by adding a rule, but must be added using the regex-style match for specificity which could catch people out and can be hard to debug if you don't have the source.

eg (the `~` marks this as regex-style):

```nginx
location ~ /\.well-known/webfinger {
	try_files $uri /index.php$is_args$args;
}
```

roborourke:
Should we just bake that into the main nginx config?

ferschubert-hm:
- Ubuntu 22 not supported, this is relevant to scan current Vantage Dashboard
```
AGENT_MANIFEST_URL: https://s3.eu-central-1.amazonaws.com/aws-agent.eu-central-1/linux/releases/1.1.1921.0/AGENT_MANIFEST
AGENT_VERSION_URL:  https://s3.eu-central-1.amazonaws.com/aws-agent.eu-central-1/linux/releases/1.1.1921.0/VERSION
PUBKEY_FILE: /tmp/awsagent.g812qMSG/inspector.gpg
Validated AGENT_MANIFEST signature with: Primary key fingerprint: DDA0 D4C5 10AE 3C20 6F46  6DC0 2474 0960 5836 0418
PUBKEY_FILE: /tmp/awsagent.g812qMSG/inspector.gpg
Validated VERSION signature with: Primary key fingerprint: DDA0 D4C5 10AE 3C20 6F46  6DC0 2474 0960 5836 0418
using legacy VERSION file
Failed to find an inspector agent package for this OS:ubuntu-6.5..
Download command failed with exist status code: 3
```

PR #952, opened 2025-04-15
Document how to allow access to blocked URLs

mikelittle:
Adds documentation with an example of how to override the default blocking of certain URLs in nginx.

Fixes https://github.com/humanmade/altis-cloud/issues/712

### DIFF
--- a/docs/nginx-configuration.md
+++ b/docs/nginx-configuration.md
@@ -69,3 +69,18 @@ location ~* config.local.yaml {
     return 404;
 }
 ```
+
+If you wanted to allow a normally blocked location to be accessed, you could use a `location` block to override the default. For
+example, Altis blocks all urls beginning with a `.` as these are often sensitive files you do not want to be accessed.
+
+But one use case might be to allow access to specific `.well-known` urls. For example, to allow `Webfinger` requests you would
+add the following location block.
+
+```nginxconf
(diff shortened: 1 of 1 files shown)


docs/nginx-configuration.md
```diff
@@ -69,3 +69,18 @@ location ~* config.local.yaml {
     return 404;
 }
 ```
+
+If you wanted to allow a normally blocked location to be accessed, you could use a `location` block to override the default. For
+example, Altis blocks all urls beginning with a `.` as these are often sensitive files you do not want to be accessed.
+
+But one use case might be to allow access to specific `.well-known` urls. For example, to allow `Webfinger` requests you would
+add the following location block.
+
```
ferschubert-hm, line 78: Would be good to keep a standard as on line 54 and add in which file this should be added, in this case is also `.config/nginx-additions.conf` (if so then also add the file to the example to block specific files line 65).

ferschubert-hm (approved):
LGTM, tested on my local env, working as expected. 

Just left a comment to add the file which this should be added like we did on line 54.

docs/nginx-configuration.md
mikelittle: I added that. Thanks for the suggestion.

PR #953, opened 2025-04-15
adds a Altis usage page to documentation

owaincuvelier:


rmccue:
> Is there a link to the "Altis data processing Addendum"?

https://www.altis-dxp.com/policies/dpa/ is the permanent link to it

Issue #948, opened 2025-03-31
Add "dclid" as a default batcache cache key exclusion


owaincuvelier:
for ignored query string parameters in batcache we currently have these defaults set:
"utm_campaign",
"utm_medium",
"utm_source",
"utm_content",
"utm_term",
"mc_cid",
"mc_eid",
"fbclid",
"_ga"

Another common one is `dclid` which will bust the batcache for marketing campaigns that bust this.

owaincuvelier:
https://github.com/humanmade/altis-cloud/blob/333fad313383ebcedc0501616ab3d00037e3fbe0/load.php#L32

ferschubert-hm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Altis-6284e0dfb386b00c83981426/issues/gh/humanmade/altis-cloud/948?planning-poker) with Zenhub @wisyhambolu @vladislavhmn

rmccue:
Should we consider adding this to the CDN as well?

wisyhambolu:
@filter182 can you please connect the pull request to this issue and move the issue to the review column?

Also, have you tested the changes to confirm they work as expected?

Lastly, don't forget the second acceptance criteria which is for the documentation update
> Update the documentation - https://docs.altis-dxp.com/cloud/page-caching/#cache-key-calculation

filter182:
@wisyhambolu  sorry missed the comment, I created the PR for the documentation, sorry I left the changes and draft. 

How can I manually test this to make sure this actually works as intended? @mikelittle mentioned in the PR that he already manually tested it but for learning purposes I would like to know how can perform this in the future.

Thanks guys! sorry for the delay.

filter182:
@rmccue it doesn't seem we do this by default which would be a great idea, I found a couple of envs that we have a custom config for that on `infrastructure-config` repo like [Rappler](https://github.com/humanmade/infrastructure-config/blob/f6e040792645258e3410851f2f856668acd5a25d/terraform/apps/rappler/rappler-shared/main.tf#L44) and [RTE](https://github.com/humanmade/infrastructure-config/blob/f6e040792645258e3410851f2f856668acd5a25d/terraform/apps/recipetineats/recipetineats-shared/main.tf#L42)

filter182:
@wisyhambolu sure, I'll create it

ferschubert-hm:
Create backports to the supported releases: https://docs.altis-dxp.com/releases/

[...]

PR #961, opened 2025-05-02
Adding dclid, gclid, and twclid to the default exclusion list

filter182:
- Delivers https://github.com/humanmade/altis-cloud/issues/948

In the above issue, we are adding three more parameters (dclid, gclid, and twclid) to the cache key exclusion list to avoid cache sharding

### DIFF
--- a/load.php
+++ b/load.php
@@ -38,9 +38,9 @@ add_action( 'altis.modules.init', function () {
 				'mc_eid',
 				'fbclid',
 				'_ga',
-		                'dclid',
-		                'gclid',
-		                'twclid',
+				'dclid',
+				'gclid',
+				'twclid',
 			],
 			'unique-headers' => [
 				'cloudfront-viewer-country',
--- a/load.php
+++ b/load.php
@@ -38,6 +38,9 @@ add_action( 'altis.modules.init', function () {
 				'mc_eid',
 				'fbclid',
 				'_ga',
+		                'dclid',
+		                'gclid',
+		                'twclid',
 			],
 			'unique-headers' => [
 				'cloudfront-viewer-country',


ferschubert-hm:
Can you please check and fix the linter notes.

filter182:
> Can you please check and fix the linter notes.

@ferschubert-hm Fixed the indentation, the rest of the errors were already there before my changes

mikelittle (changes requested):
I tested this manually and confirmed the changes work. But as noted on the ticket, the docs need updating.
Ideally, as a minimum manual testing should be at least documented on the ticket/this pr

filter182:
> I tested this manually and confirmed the changes work. But as noted on the ticket, the docs need updating. Ideally, as a minimum manual testing should be at least documented on the ticket/this pr

Hey @mikelittle! thanks for testing this, sorry for the delay. I created the PR to update the documentation https://github.com/humanmade/altis-cloud/pull/962

question, how can I manually test this?

mikelittle:
Hi @filter182 , here's what I did to test this.

First confirm the problem exists without your change.
Assuming an up to date checked out `product-dev` repo, enable bat cache locally: in composer.json
```json
"extra": {
 "altis": {
      "modules": {
        "cloud" : {
          "batcache": true
        },...
```

Then without the change in place (that is, with the packages/cloud on `master`) use something like `curl -i "https://product-dev/sample-page/" and look for the  `x-batcache` header. The first time it will say "MISS"
Then run the same curl command. This time, the `batcache` header should say "HIT".

Next try it with `curl -i "https://product-dev/sample-page/?gclid=`123abc", you will get "x-batchache: MISS"
Running it again will get a HIT. 
Change the `gclid` value and you will get a miss again.
This shows that `batcache` is not ignoring this query parameter.

Then switch to your PR branch, and run the same tests.
```
cd packages/cloud/
git switch jc-948-altis-cloud
cd ../..
```

This time, the first use of `gclid` should get you a HIT and the same response as without the parameter (you can tell by the timestamp info at the bottom of the page)
Then changing the parameter will still get you a HIT
You can add 1, 2, or all three parameters, with any values (and change them) and you will still keep getting HIT (though it will expire every 5 minutes)

That shows `batcache` is now ignoring those three parameters.

Does that make sense?

filter182:
> Hi @filter182, here's what I did to test this.
> 
> First confirm the problem exists without your change. Assuming an up-to-date checked out `product-dev` repo, enable bat cache locally: in `composer.json`
> 
> ```json
> "extra": {
>  "altis": {
>       "modules": {
>         "cloud" : {
>           "batcache": true
>         },...
> ```
> 
> Then without the change in place (that is, with the packages/cloud on `master`) use something like `curl -i "https://product-dev/sample-page/"` and look for the `x-batcache` header. The first time, it will say "MISS" Then run the same curl command. This time, the `batcache` header should say "HIT".
> 
> Next try it with `curl -i "https://product-dev/sample-page/?gclid=123abc"`, you will get "x-batcache: MISS" Running it again will get a HIT. Change the `gclid` value and you will get a miss again. This shows that `batcache` is **not** ignoring this query parameter.
> 
> Then switch to your PR branch, and run the same tests.
> 
> ```
> cd packages/cloud/
> git switch jc-948-altis-cloud
> cd ../..
> ```
> 
> This time, the first use of `gclid` should get you a HIT and the same response as without the parameter (you can tell by the timestamp info at the bottom of the page) Then changing the parameter will still get you a HIT You can add 1, 2, or all three parameters, with any values (and change them) and you will still keep getting HIT (though it will expire every 5 minutes)
> 
> That shows `batcache` is now ignoring those three parameters.
> 
> Does that make sense?

@mikelittle thanks for detailed explanation! I'll test this out.

wisyhambolu (approved):
Changes look good 👍

PR #983, opened 2025-06-30
Order PHP information from newest to oldest.

johnbillion:
Bugbear of mine. This sort of information reads better when displayed newest to oldest.


mikelittle (approved):
I've no objections to this change. Thanks.

PR #994, opened 2025-08-01
Add X-Ray trace to error messages

joehoyle:
All PHP error messages will now be linked to an X-ray trace when X-Ray is loaded. We also add the CLI command if instead it's coming from the CLI.

See https://github.com/humanmade/vantage/issues/239


inc/namespace.php
```diff
@@ -1225,3 +1227,18 @@ function register_trace_qm_panel_menu_item( array $menu ) : array {
...
+ *
+ * This requires the https://github.com/humanmade/error-message-format/ PHP
+ * extension to be installed. If it's not, this will have no effect.
+ *
+ * @return void
+ */
+function set_php_error_message_format() : void {
+	if ( function_exists( 'HM\\Platform\\XRay\\get_root_trace_id' ) ) {
+		$trace_id = XRay\get_root_trace_id();
+		ini_set( 'error_message_format', '{message} (Trace ID: ' . $trace_id . ')' );
```
rmccue, line 1242: Can this return null/etc? (eg locally if there's no ID header)

joehoyle: No, this always returns a string. If no trace ID is in the HTTP headers, it will generate one

rmccue (approved):

PR #1035, opened 2025-10-01
Clarify which bots are not counted for usage

rmccue:
(I used the edit button on the page so this is on the v24 branch, probably needs cherrypicking to main.)


owaincuvelier (approved):

mikelittle (approved):
Nice clarifications.

PR #1037, opened 2025-10-06
Add docs for Log Shipping

jerico:
Part of https://github.com/humanmade/product-dev/issues/1859

### DIFF
--- a/docs/log-shipping.md
+++ b/docs/log-shipping.md
@@ -1,0 +1,100 @@
+# Log Shipping
+
+Log Shipping sends a copy of your access logs to an S3 bucket you own on a regular basis.
+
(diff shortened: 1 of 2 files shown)


docs/log-shipping.md
```diff
@@ -0,0 +1,62 @@
...
+        "AWS": "arn:aws:iam::577418818413:role/<environment-name>-log-shipping-s3-events"
+      },
+      "Action": [
+        "s3:PutObject"
+      ],
+      "Resource": "arn:aws:s3:::<DEST_BUCKET>/<environment-name>/cloudfront/*"
+    }
+  ]
+}
+```
```
wisyhambolu, line 50: @jerico I don't think we need to share all this in our public docs, instead of this we can just point users to contact Altis support and we can provide them with them with the required configuration via the support ticket.

PR #1070, opened 2025-11-17
Add --install for nvm

rmccue:
This avoids an edge-case bug on a new repository. See https://github.com/nvm-sh/nvm/issues/1985


mikelittle (approved):
Looks good to me.

It looks like it took a lot of work to figure out the edge case on that original ticket.

mikelittle:
Do we want to backport this ti supported versions?

PR #1161, opened 2026-04-17
Add search-replace on sync docs

jerico:
Part of https://github.com/humanmade/vantage/issues/989

### DIFF
--- a/docs/dashboard/search-replace.md
+++ b/docs/dashboard/search-replace.md
@@ -1,8 +1,9 @@
 # Search and Replace on Sync
 
-During a database sync, Altis can automatically perform string replacements on imported data. Mappings are applied in-transit
-before data is stored in the destination database. This allows environment-specific values, such as domain names, to be updated for
-the target environment as part of the sync process.
+During a database sync, Altis can automatically perform string replacements on
+imported data. Mappings are applied in transit before data is stored in the
+destination database. This allows environment-specific values, such as domain
+names, to be updated for the target environment as part of the sync process.
 
 Mappings are fetched from `composer.json` at the currently deployed SHA.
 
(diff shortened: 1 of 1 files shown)


joehoyle (approved):

docs/dashboard/search-replace.md
```diff
@@ -0,0 +1,52 @@
...
+destination database. This allows environment-specific values, such as domain
+names, to be updated for the target environment as part of the sync process.
+
+Mappings are fetched from `composer.json` at the currently deployed SHA.
+
+## Configuration
+
+Configure mappings under `extra.altis.cloud.search-replace` in `composer.json`.
+
+Each top-level key should be the exact ID of the environment you are syncing
```
rmccue: I think we should emphasise this:

```suggestion
Each top-level key should be the **exact** ID of the environment you are syncing
```

docs/dashboard/search-replace.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+In this example, if a database is synced into `example-stag-01`, every occurrence of
+`example.com` will be replaced with `example-stag.altis.cloud`.
+
+Likewise, if a database is synced into `example-dev-01`, both `example.com` and
+`www.example.com` will be replaced with `example-dev.altis.cloud`.
+
+## Notes
```
rmccue: I'd either move this to the top as a summary, or flesh it out a bit with details.

rmccue (dismissed):
Should this be a page about synching instead of just search-replace? I'm not sure if we've got any docs on that currently, but feels like it should be colocated

mikelittle (changes requested):
Agree with Ryan the info in notes is important to come before the example, so someone doesn't stop at the example and not read the rest.

jerico:
Environment sync/import/export have it's own page already https://docs.altis-dxp.com/cloud/dashboard/import-export/ but needs to be updated in this issue https://github.com/humanmade/product-dev/issues/1861?issue=humanmade%7Cproduct-dev%7C1976

mikelittle (approved):
Looks good to me

PR #1200, opened 2026-06-29
Pin s3 uploads version

mikelittle:
This change pins `humanmade/s3-uploads` to version 3.0.12.
Allowing composer to pull in a higher version (which `s3-uploads` 3.0.13 does) is incompatible with the `aws-sdk-php` version pinned in `altis/core`. Which, in turn, would break Local Server uploads if allowed to increase.

Addresses issue https://github.com/humanmade/product-dev/pull/2187


joehoyle (approved):
@mikelittle are we tracking this mismatch to later fix up?

PR #1223, opened 2026-07-29
Cross-link new customization options

rmccue:
We've got a growing body of text under this section, so let's cross-link it a bit better. Since we originally wrote the customizations section, there's now a lot more that users can do themselves, so this fleshes out those docs a bit.

### DIFF
--- a/docs/WAF/README.md
+++ b/docs/WAF/README.md
@@ -7,6 +7,9 @@ and [AWS Shield Advanced](https://aws.amazon.com/shield/).
 Our firewall has protected customers against active attacks, including HTTP request floods up to 1 million requests per second,
 sustained over hours.
 
+WAF behaviour is [customizable](#firewall-customizations) including self-service changes within the Altis Dashboard, and custom
+rulesets available as an add-on.
+
 ## Protection against exploits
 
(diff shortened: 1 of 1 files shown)


docs/WAF/README.md
```diff
@@ -88,18 +90,26 @@ Once an alert is validated, engineers begin our incident management process:
...
-as [custom nginx configuration](./nginx-configuration.md).
+Our firewall is maintained as standard across our platform, and for most customers, you'll never need to worry about your site. For
+those cases when you do, we provide a variety of customizations you can adjust:
+
+* Control incoming traffic by category or signal with [Traffic Management](traffic-management.md)
+* Block individual user-agents or exclude them from rate limits with [User-Agent Controls](ua-blocking.md)
+* Manage access by IP address or range with [IP Access Control](access-control.md)
+
+For more advanced control, many additional controls including blocking routes and behaviours can be implemented via
+[custom nginx configuration](../nginx-configuration.md). This allows writing arbitrary nginx rules, however operates *after* the
```
mikelittle:
```suggestion
[custom nginx configuration](../nginx-configuration.md). This allows writing arbitrary nginx rules, however these rules operate *after* the
```

mikelittle (changes requested):
Just one minor change. Otherwise looking good.

mikelittle (approved):

PR #1226, opened 2026-08-17
Explicitly note different key

rmccue:
The use of the top-level cloud key for configuration can be confusing, as we also have a cloud module. This clarifies that it is intentionally different.


mikelittle (approved):
Great,. Thankyou.

